Issue #9, opened 2024-12-06
Allow users to configure max zoom


blms:
The Atlas client has requested that zoom be made less "aggressive"—one way to mitigate this would be a `maxZoom` config option that is passed to `boundingBoxOptions`, with the default remaining at 14.

Issue #8, opened 2024-12-06
Allow users to disable zoom on place detail selection


blms:
The Atlas client would like to mitigate some of the zoom-based context loss when navigating to a particular place on the map. 

My proposal is a configuration option to disable zoom onto a single entry when clicking into a place detail page. That way, the context of the other search results is maintained.

PR #11, opened 2024-12-17
Allow configuring max zoom and individual place zoom (#8, #9)

blms:
## In this PR

Per #8:
- Add a `search.zoom_to_place` boolean config option to allow disabling `fitBoundingBox` on the place detail page. In other words, allow users to disable zooming to an individual place on selection. Searches/filters narrowing the results will still zoom as before.

Per #9:
- Add a `search.max_zoom` numeric config option for adjusting the max zoom of the map.

## Notes

- Something I noticed while setting up my instance: the readme links to https://github.com/performant-software/core-data-tina-cms-content but that gives me a 404. Maybe it's a private repo? If it's been removed, we should update the readme.

### DIFF
--- a/public/config.json
+++ b/public/config.json
@@ -20,7 +20,10 @@
     "cluster_radius": 6,
     "result_card": {
       "title": "name"
-    }
+    },
+    "timeline": false,
+    "zoom_to_place": false,
+    "max_zoom": 8
   },
   "typesense": {
     "host": "example.typesense.com",
--- a/public/config.json
+++ b/public/config.json
@@ -22,8 +22,8 @@
       "title": "name"
     },
     "timeline": false,
-    "zoom_to_place": false,
-    "max_zoom": 8
+    "zoom_to_place": true,
+    "max_zoom": 14
   },
   "typesense": {
     "host": "example.typesense.com",
--- a/src/apps/search/Place.tsx
+++ b/src/apps/search/Place.tsx
@@ -14,7 +14,7 @@ import {
   usePlacesService,
 } from '@performant-software/core-data';
 import { LocationMarkers } from '@performant-software/geospatial';
-import { type AnnotationPage, useCurrentRoute, useNavigate } from '@peripleo/peripleo';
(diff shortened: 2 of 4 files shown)


src/apps/search/Place.tsx
```diff
@@ -132,6 +133,7 @@ const Place = () => {
         <LocationMarkers
           animate
           boundingBoxOptions={bboxOptions}
+          fitBoundingBox={Object.hasOwn(config.search, "zoom_to_place") ? config.search.zoom_to_place : true}
```
dleadbetter, line 136: The `underscore` library has a nice [get](https://underscorejs.org/#get) function that we could probably use here to avoid a ternary operator. Single quotes on string values.

public/config.json
```diff
@@ -20,7 +20,10 @@
     "cluster_radius": 6,
     "result_card": {
       "title": "name"
-    }
+    },
+    "timeline": false,
+    "zoom_to_place": false,
+    "max_zoom": 8
```
dleadbetter, line 26: Let's keep the `max_zoom` property as the default of "14".

dleadbetter (commented):
Just a couple of minor things, but looks good! Could we also update the table in the [Getting Started](https://github.com/performant-software/core-data-places/tree/develop?tab=readme-ov-file#getting-started) section of the `README.md` to include the new configuration properties?

> Something I noticed while setting up my instance: the readme links to https://github.com/performant-software/core-data-tina-cms-content but that gives me a 404. Maybe it's a private repo? If it's been removed, we should update the readme.

That link appears to have been removed in the `develop` branch, but if you're still seeing it there, let me know.

src/apps/search/Place.tsx
```diff
@@ -132,6 +133,7 @@ const Place = () => {
         <LocationMarkers
           animate
           boundingBoxOptions={bboxOptions}
+          fitBoundingBox={_.get(config.search, "zoom_to_place", true)}
```
dleadbetter, line 136: Being nit-picky, but single quotes around the string value here.

blms: Oops! 😅

dleadbetter (approved):

PR #14, opened 2024-12-19
Add default_sort and bump dependencies (#13)

blms:
## In this PR

Per #13:
- Add the `default_sort` option to the readme and example config.
- Bump required versions of react components libraries

Also:
- Bump netlify CLI, while we're here ([no breaking changes](https://github.com/netlify/cli/releases))


dleadbetter (approved):
Looks great!

Issue #7, opened 2024-12-06
Navigating between place detail pages causes data to be appended rather than replaced


blms:
When navigating from one place detail page to another using the map pins, data will be appended to the detail section rather than replacing its contents.

https://github.com/user-attachments/assets/73c4e33b-dd2e-412f-987b-fdb4270f0c6d

PR #16, opened 2024-12-19
Prevent key collision, resolve component duplication (#7)

blms:
## In this PR

Per #7:
- Fix a bug where place detail data would be appended rather than replaced, by preventing a key collision.

## Notes

Two adjacent components `PlaceDetails` and `RelatedRecords` shared a `key` attribute. For whatever reason, React decided that this meant it should append a new `PlaceDetails` whenever that key changed, while keeping the old one rendered. Resolving the collision also stopped this behavior.

### DIFF
--- a/src/apps/search/Place.tsx
+++ b/src/apps/search/Place.tsx
@@ -121,7 +121,7 @@ const Place = () => {
           />
         )}
         <RelatedRecords
-          key={uuid}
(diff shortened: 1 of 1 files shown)


dleadbetter (approved):
Interesting, I was unaware of this React behavior, but good catch 👍 .

PR #19, opened 2024-12-27
Initial pass at upgrading astro and tina to latest

ajolipa:
### In this PR
Upgrades to Astro v5.1.1 and TinaCMS v2.5.2, along with related dependencies. Also upgraded the PSS React component library packages to 2.2.20 while I was at it (this shouldn't be a breaking change).

The purpose of this PR is mainly to create a deploy preview to test the upgrades; locally nothing seems to have broken as far as I can tell so far.

UPDATE: Now updating to TinaCMS v2.6.0 and react component library 2.2.22.

### How to test
Ideally making these upgrades will not lead to any noticeable changes in the sites, so all functionality should be tested to make sure there weren't breaking changes we didn't catch:
- Map search, filtering, places detail panel
- Posts, including embedded map components
- Path viewer navigation and map functionality
- TinaCMS -- should be able to update and create content, including selecting places from the associated Core Data project to embed in posts or add to paths, adding images, etc
- Site should function smoothly in general (no broken links, loading errors, etc.)


dleadbetter (commented):
This looks good to me. Is there a GitHub issue we can link?

dleadbetter (approved):

PR #20, opened 2024-12-29
Playing around with Keycloak auth

ajolipa:
### In this PR
~This is not close to ready to be merged, just making a PR so I can play with the deploy previews on Netlify.~

Implements KeyCloak auth for TinaCMS


auth.config.ts
```diff
@@ -0,0 +1,15 @@
...
+import dotenv from 'dotenv';
+
+dotenv.config();
+
+export default defineConfig({
+  providers: [
+    Keycloak({
+      clientId: process.env.AUTH_KEYCLOAK_ID || 'tinacms',
+      clientSecret: process.env.AUTH_KEYCLOAK_SECRET || '',
+      issuer: process.env.AUTH_KEYCLOAK_ISSUER || 'https://keycloak.archivengine.com/realms/GBoF'
```
dleadbetter, line 12: Do we want to hard-code `https://keycloak.archivengine.com/realms/GBoF` here?

ajolipa: Ah, probably not, good call.

Issue #23, opened 2025-01-08
Add "Show more" option to facet lists


dleadbetter:
By default, the groups of facets only display the first 10 options. We should add a "Show more" option to allow users to view the full list of facet values.

PR #24, opened 2025-01-09
CDP #23 - Facets show more

dleadbetter:
This pull request adds the ability to click a "Show More" button for each facet list to view the entire list of facet values. It also adds the ability to include a `typesense.facets.include` or `typesense.facets.exclude` options to control which facets display on the modal. Lastly, this pull request adds some missing `defaultValue` attributes to i18n.json.

Screenshot 2025-01-09 at 11 51 38 AM


blms (approved):
Looks good to me! I think these options will help a lot with reducing the clutter of the facets. I also checked the react-components updates and I like how you're handling the include/exclude logic. 👍

PR #25, opened 2025-01-09
Handling for local config

camdendotlol:
# Summary

This PR adds handling to the `init` function to allow developers to point to a local file in `CONFIG_URL`.

When `TINA_PUBLIC_IS_LOCAL` is set to true, `init` now uses `fs.readFileSync` to load the config file. When `TINA_PUBLIC_IS_LOCAL` is false, the current behavior (a `fetch` request) is used.


build.config.mjs
```diff
@@ -1,13 +1,29 @@
-import dotenv from 'dotenv';
```
dleadbetter, line 1: Let's use single quotes to be consistent with the rest of the project.

dleadbetter (commented):
This is nice, I like this idea a lot. I find myself often overwriting the `public/config.json` when building.

Instead of pulling the config from the local file system path when `TINA_PUBLIC_IS_LOCAL` is "true", would it be trivial for developers to simply paste the contents of the desired config into `/public/config.json`? This way we could decouple the "local" config from whether or not we're using a local Tina instance. I think the business logic could be as follows:

- If a value exists for `CONFIG_URL`, use the `fetch` API to pull in the config
- If _no_ value exists for `CONFIG_URL`, use the existing `public/config.json`

dleadbetter:
Also, let's add an issue to GitHub describing this change.

build.config.mjs
camdendotlol: Is there a linting/formatting config for this project? I noticed my default Prettier was reformatting things but I don't see a project-specific config for either it or ESLint. I see `prettier-plugin-astro` in the dependencies but not Prettier itself.

I can just turn off format-on-save and try to manually match my formatting if needed.

dleadbetter: I don't believe there is a liniting/formatting config yet, I'll add an issue to GitHub.

dleadbetter (approved):

PR #29, opened 2025-01-13
Optional polygon search view

camdendotlol:
# Summary

This PR adds a new, optional `polygon` option in the `search` object in `config.json`. Enabling it will tell the search components to display polygons in the map search view, when available, instead of points.

It also adds a couple bits of logic:

* disable clustering when polygons are enabled, as `maplibre-gl` does not support clustering with polygons: https://github.com/mapbox/mapbox-gl-js/issues/3419#issuecomment-603918106
* use Typesense's `exclude_fields` param to exclude the `geometry` field when polygons are disabled, making the search responses smaller


src/apps/search/TypesenseSearch.tsx
```diff
@@ -46,9 +46,23 @@ const SearchProvider = (props: { children: ReactNode }) => {
 
 const TypesenseSearch = (props: { children: ReactNode }) => {
   const config = useRuntimeConfig<any>();
+
+  const typesenseConfig = useMemo(() => {
+    // Exclude the geometry field (which contains polygons) when polygon map results
+    // are disabled. This saves bandwidth and loading time caused by unnecessarily
+    // loading extremely large polygons (i.e. > 1MB in some cases).
+    if (!config.search.polygons) {
```
dleadbetter, line 54: I'm not sure we need to do this automatically. I think it can be up to the developer to configure these settings appropriately.

dleadbetter (approved):

Issue #33, opened 2025-01-17
Update "core-data" package to be model independent


dleadbetter:
Update the `core-data` package such that it does not rely on place records being returned from Typesense/CoreData API.

dleadbetter:
## Testing Note

Previously, this site was only capable of presenting Place records from Core Data. We've since updated it such that it can be configured to present _anything_ coming out of Core Data. For this issue, it may be worth doing some general testing with the three different staging sites we have up that use different types of records:

#### Places (PSS Scavenger Hunt)
https://pss-scavenger-hunt.netlify.app/en/

#### Organizations (Ground Beneath Our Feet - C19 NYC)
https://gbof-c19nyc-staging.netlify.app/en/

#### Events (Pan-African Data)
https://padp-staging.netlify.app/en

PR #43, opened 2025-01-30
CDP #33 - Core data models

dleadbetter:
This pull request refactors the `core-data-places` repository to work with any type of model coming out of Typesense/Core Data. Previously, it was required that only place records be indexed.

- Reorganizes the `config.json` options (💥 breaking change!)
- Adds model specific side panels for each model type
- Adds `UserDefinedFieldView` component for user-defined fields rendering

Below are some examples of configuring the site to work with different types on models.

#### Search results
Screenshot 2025-01-30 at 9 52 18 AM

#### Side panel
Screenshot 2025-01-30 at 9 52 35 AM

#### Search results
Screenshot 2025-01-30 at 9 53 47 AM

#### Side Panel
Screenshot 2025-01-30 at 9 55 31 AM

#### Related records
Screenshot 2025-01-30 at 9 55 38 AM


src/apps/search/panels/Place.tsx
```diff
@@ -0,0 +1,76 @@
+import RecordPanel from '@apps/search/panels/RecordPanel';
+import { useTranslations } from '@i18n/client';
+import {
+  CoreData as CoreDataUtils,
+  PlaceLayersSelector,
+  useLoader,
+  usePlacesService
+} from '@performant-software/core-data';
+import { useCurrentRoute, useNavigate } from '@peripleo/peripleo';
+import { getPreviewImage } from '@utils/Media';
```
ajolipa, line 10: Should be `@utils/media` I think; this seems to be what made the deploy preview fail.

ajolipa (approved):
This looks good to me other than the one typo (see comment). Should we update the config file for pss-scavenger-hunt to the new format so we can test with the deploy preview?

package-lock.json
camdendotlol, line 1: Why wasn't `package-lock.json` checked in prior to this PR?

camdendotlol (approved):

src/apps/search/panels/Place.tsx
dleadbetter: Good catch, it was also an issue in `BasePanel`.

package-lock.json
dleadbetter: Good question. Looks like it was removed in a previous [PR](https://github.com/performant-software/core-data-places/pull/19/files) and went unnoticed. @ajolipa - Any thoughts on why that may have been removed?

ajolipa: Oh, yeah I think that was my fault; it was deleted at some point in the process of rebasing and I never re-committed it. That's my bad, sorry for confusion!

PR #45, opened 2025-02-06
Update search.max_zoom to map.max_zoom

blms:
## In this PR

Cleanup: code and example config still contained references to `search.max_zoom` when that property was changed to `map.max_zoom` in #43.


dleadbetter (approved):
Good catch 👍 .

PR #40, opened 2025-01-28
Rb astro cd endpoints

ajolipa:
### In this PR
This is a draft PR for the purpose of triggering a deploy preview on Netlify to test the static pages workflow. Will update this when more testing has been done.

PR #47, opened 2025-02-10
Adding `[lang]/[model]/[slug].astro` pages for featured models

ajolipa:
### In this PR
Adds static pages for each record in the whitelisted model types, with each possible model type having its own Astro component that can be modified individually for different designs per model. The static detail pages pull their data from one of two places:
- In SSG mode where the Astro content layer is being used, they use the Astro content API to fetch the record's entry from the datastore.
- In SSR mode, they fetch the data from the Astro endpoint `api/[model]/[slug]`, which in turn uses per-model helper functions to fetch the data from the Core Data public API at request time.

This should be merged after PR #40


src/pages/api/[model]/[slug].json.ts
Comment (jamiefolsom) [inline src/pages/api/[model]/[slug].json.ts:1]: @dleadbetter this might be unnecessary or infeasible, but should we (and actually can we) version this API route, so that if/when we change it, existing sites can decide when to opt in to a newer version? So the routes would be something like: `src/pages/api/[version]/[model]/[slug].json.ts` -- @ajolipa is that possible?

Comment (ajolipa) [reply src/pages/api/[model]/[slug].json.ts:1]: (Sounds like this question is really about PR #40 ?) Hmm... I think that could make sense, and it would be nice to label it in accordance with the version of the public API that it's getting payloads from. It would certainly be possible to add that parameter to the path.

src/pages/[lang]/[model]/[slug].astro
```diff
@@ -0,0 +1,58 @@
...
+const lang = Astro.currentLocale;
+
+let data: any;
+
+if (prerender) {
+  // @ts-ignore
+  const entryResponse = await getEntry(model, slug);
+  data = entryResponse.data;
+} else {
+  const apiUrl = new URL(`/api/${model}/${slug}.json`, Astro.url.origin);
```
Comment (jamiefolsom) [inline src/pages/[lang]/[model]/[slug].astro:27]: if we do version the API, should this be `/api/${version}/${model}/${slug}.json`?

dleadbetter:
Closing in favor of #143.

PR #49, opened 2025-02-12
Fixing `package-lock.json` file

ajolipa:
### In this PR
Attempt to correct the broken state that `package-lock.json` ended up in at some point during a rebase, oops!


blms (approved):
Success! I did notice that it seems to have some substantial changes from two commits prior, but maybe that's fine since it all still installs and runs.

My fresh install from commit `ebacfbaf86b48205c710550ef56f972d5e8ee204`:
```
added 3944 packages, and audited 3947 packages in 2m
```
And then installing from this branch afterwards:

```
added 439 packages, removed 663 packages, changed 337 packages, and audited 3723 packages in 35s
```

dleadbetter (approved):

Issue #44, opened 2025-01-30
Allow map layer selection from Map Library


dleadbetter:
### Testing notes

On the [GBoF staging site admin](https://gbof-c19nyc-staging.netlify.app/admin):
- In "Settings", under "Layers", click the + sign to add a new layer
- Edit the new layer, and you should see an option called "Georeference" under layer type
- Selecting that should bring up the option to choose from the Map Library
- If you choose a layer from the map library, it should populate the URL field
- It should not be possible to uncheck "Overlay" (all georeferenced layers are overlays)
- Name and save the new layer record, and make sure the overall settings are saved as well (this might happen automatically).
- Return to the map on the front page. The new layer should show up in the layers list by clicking the layers icon on the right side (n.b. requires a redeloy)

---

When editing website config via TinaCMS, under the "Layers" section, allow selection of a URL from the Map Library. The map library should only be available when "georeference" is selected as a layer type.

1. Add an additional environment variable for `MAP_LIBRARY_URL`
  a. I've setup some mock JSON data [here](https://core-data-map-library.s3.us-east-1.amazonaws.com/library.json) for now that we can use to build this functionality

2. We should add "georeference" as an option in the layer types

<img width="1142" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBM3lwQ0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--0c48c7037ba0dc74318dcd7dae819c5d40017111/Screenshot%202025-01-30%20at%202.59.25%E2%80%AFPM.png" alt="Screenshot 2025-01-30 at 2.59.25 PM.png" />

3. When "georeference" is selected as layer type, users should have the ability to directly enter a URL or select from the Map Library (which will also store a URL under-the-hood)
  a. Perhaps we have separate tabs for "URL" and "Map Library"
  b. Use `MAP_LIBRARY_URL` environment variable to get the list of options

4. When a "georeference" layer is added, the "overlay" option must me set to "true"

jamiefolsom:
@blms here's another map to add to the library when you get to the point of adding more maps.

https://digitalcollections.nypl.org/items/f5a403f0-d044-0138-6145-1ff6d3da4cc7

PR #51, opened 2025-02-14
Add map library functionality to Tina (#44)

blms:
## In this PR

Per #44:
- Add an env variable `TINA_PUBLIC_MAP_LIBRARY_URL` to allow users to specify a map library
- Add map layer type "georeference"
- If "georeference" selected and map library url set, fetch data from map library url and show dropdown with options
   - On selecting an option, populate the `url` field with that layer's url
   - Also, force `overlay` boolean to be `true` and disable the input

https://github.com/user-attachments/assets/a34c9339-acc8-4e52-80e7-dfe58d7405e6

Screenshot 2025-02-14 at 3 20 58 PM


ajolipa (approved):
This rules.

dleadbetter (commented):
Looks great! Any thoughts on the build error?

blms:
Just noticed a mistake in the conditional for showing the map library dropdown when editing a layer, so hopefully that should rebuild successfully!

dleadbetter (approved):

Issue #46, opened 2025-02-10
Layout design


dleadbetter:
Re-design layout.

<img width="1085" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd1RNQ0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--443a43631a0e2e9f4028d5520a34b87ab036b21b/Screenshot%202025-02-10%20at%209.10.33%E2%80%AFAM.png" alt="Screenshot 2025-02-10 at 9.10.33 AM.png" />

dleadbetter:
## Testing Notes

This is a complete re-design of the previous layout. We can use this issue as a general smoke test for the new layout. Please open new issues for any bugs found.

PR #59, opened 2025-02-19
CDP #46 - Layout redesign

dleadbetter:
This pull request re-designs the map layout for the `core-data-places` site.

**Note:** This PR is dependent on a new release of `react-components`.

## Default View
By default, the search results will display in a panel on the left side of the screen and the map is visible.

Screenshot 2025-02-18 at 7 45 15 AM

## Filters Panel
The filters/facets panel can be toggled open or closed and will display to the left of the search results panel.

Screenshot 2025-02-18 at 7 45 21 AM

## Record Detail Panel
When clicking on a record in the search results list, table, or map geometry/point a detail panel for the selected record will open on the right side of the screen.

Screenshot 2025-02-18 at 7 45 31 AM

## Table View
Instead of the search results list view, a table view is also toggleable using the List/Table buttons in the header.

Screenshot 2025-02-18 at 7 45 42 AM

## Table View with Detail Panel
When a detail panel is opened at the same time as the table view, the table will shrink in width to accommodate the detail panel.

Screenshot 2025-02-18 at 7 48 12 AM


public/config.json
```diff
@@ -34,7 +36,10 @@
     "api_key": "abcdefg",
     "index_name": "core_data_places",
     "query_by": "name",
-    "default_sort": "name"
+    "default_sort": "name",
+    "overrides": {
+      "geoLocationField": "coordinates"
+    }
```
blms, line 42: Should this be configurable in tina?

src/apps/search/panels/BasePanel.tsx
```diff
@@ -1,77 +1,312 @@
...
+
+    return [{
+      items: _.map(collection.items, (item) => item),
+      renderItem: (item) => (
+        <ManifestThumbnail
+          className='py-2 px-6'
+          itemCount={item.item_count}
+          name={_.first(item.label?.en)}
+          onClick={() => setManifestUrl(item.id)}
+          thumbnail={item.thumbnail}
```
blms, line 179: One thing I noticed in Atlas is that if a thumbnail failed to generate or otherwise didn't exist, the related media items would become inaccessible. It might be worthwhile to do a conditional check for `item.thumbnail` and have a backup button that can still set the manifest url to the item id.

src/apps/search/panels/BasePanel.tsx
```diff
@@ -1,77 +1,312 @@
...
 
   /**
    * Memo-izes the name.
    */
   const name = useMemo(() => (item && props.renderName && props.renderName(item)) || item?.name, [item]);
 
+  /**
+   * Transforms all the related records.
+   */
+  const relatedRecords = useMemo(() => [
```
camdendotlol, line 195: How often will this value be recomputed? There's a lot going on here and it looks like, e.g. when `people` changes, we'll be re-running tons of computations on other models that won't be any different than last time. That's probably fine if this is only being run once in practice, but otherwise I wonder if there would be a way to memoize individual object properties instead of the whole object?

src/apps/search/GeosearchFilter.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+  if (!config.search?.geosearch) {
+    return null;
+  }
+
+  return (
+    <div
+      className='text-sm flex items-center'
+    >
+      <Checkbox
```
camdendotlol, line 19: After https://github.com/performant-software/react-components/pull/405, the Checkbox component expects an `ariaLabel` prop that it can pass down to its internal `<button>`. This is true even if we're already using a `<label>` because a11y checks don't consider a `<label>` as text for a `<button>` like it would for an `<input type="checkbox" />`.

camdendotlol (changes requested):
A couple comments/questions, nothing major

public/config.json
dleadbetter: I'm not sure it's something we'd want in Tina, as this will likely be set by the developer setting up the site, not changed by the project owner. There should be [documentation](https://github.com/performant-software/core-data-places/blob/314ccdeb3d3520f848c53979f0ba6fe8b81ad56a/README.md?plain=1#L63) that links the the Typesense adapter we're using in `README.md`.

src/apps/search/panels/BasePanel.tsx
dleadbetter: 👍 good call.

src/apps/search/panels/BasePanel.tsx
dleadbetter: 👍 yeah, I think that's probably a better approach.

src/apps/search/panels/BasePanel.tsx
dleadbetter: Adding a `Button` component to render if no `thumbnail` prop is provided. We don't have a good media icon at this point, so I'm using the "info" icon as a placeholder.

Screenshot 2025-02-20 at 6 22 16 AM

camdendotlol (approved):

src/apps/search/panels/BasePanel.tsx
blms: Looks good!

Issue #52, opened 2025-02-14
Allow project owner to set default map overlays to visible in Tina


jamiefolsom:
### Testing notes

On the admin for one of the staging sites:
- Edit a map layer. If "overlay" is unchecked, "visible by default" should be greyed out
- Find a layer where "overlay" is checked, and you should then be able to check "visible by default" as well
- Save the record and return to the home page. Open the layers list using the button on the right, and the layer you set "visible by default" should be checked automatically.

---

As a project owner, I want the ability to set an overlay map layer as visible by default using TinaCMS.

~~So in the tina website setting, add a checkbox on the layer list to indicate default layer, or perhaps make the list orderable, with top item being the default.~~

PR #71, opened 2025-02-19
Add boolean for default visible overlay layers (#52)

blms:
## In this PR

Per #52:
- Add a `default` boolean property, configurable in Tina, to map layer objects, in order to make them visible by default

## Notes

- I made a new component for the input, really similar to `TinaMapLayerOverlaySwitch`, so that this field will always be disabled for non-overlay layers. I figure it's best not to let users set that true on base layers and be confused about its purpose.


dleadbetter (commented):
This looks good. I see that this will allow the user to mark a map layer a visible by default. How will that information then be passed to the map view?

blms:
@dleadbetter It's used by react-components when rendering the map layers component, which selects them on mount and then calls the `onChangeOverlays` callback for them. That was the work in this PR: https://github.com/performant-software/react-components/pull/434

dleadbetter (approved):

Issue #67, opened 2025-02-19
Update detail side panels to pull from Astro endpoints


dleadbetter:
Update the detail side panels to pull from new Astro API endpoints, rather than making requests directly to Core Data.

PR #75, opened 2025-02-21
CDP #67 - Astro endpoints for detail panels

dleadbetter:
This pull request updates the detail panel components (`@apps/search/panels`) to fetch data from the Astro endpoints instead of Core Data directly. This will allow us to fetch data at build time or request time and use the same endpoints in the Astro/React components.

This pull request also replaces the `PUBLIC_STATIC_BUILD` environment variable with the more specific `PUBLIC_BUILD_STATIC_ENDPOINTS`. If set to `true`, the `/pages/api/:model/:slug` will be built at build-time.

At request time, if `PUBLIC_BUILD_STATIC_ENDPOINTS` is set to `true`, data will be pull from the Astro content collections. If `PUBLIC_BUILD_STATIC_ENDPOINTS` is set to `false`, data will be pulled directly from Core Data API.

This pull request also removes any dependency on the `config.detail_pages` configuration option. That config option should be use to determine if a specific detail page route (e.g. `/en/places/:uuid`) is valid for the project.


src/pages/api/[model]/[slug]/[relatedModel].json.ts
```diff
@@ -1,17 +1,15 @@
...
-import config from "@config";
-import { loaderDict } from "@root/src/loaders/api";
-import { getRelation, relatedModelTypes } from "@root/src/loaders/api/helpers";
+import { getRelation } from '@loaders/api/helpers';
+import { buildStaticEndpoints, modelTypes, relatedModelTypes } from '@loaders/coreDataLoader';
+import type { APIRoute, GetStaticPaths } from 'astro';
+import { getCollection, getEntry } from 'astro:content';
 
-export const prerender = true;
+export const prerender = !buildStaticEndpoints;
```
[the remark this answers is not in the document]
Comment (dleadbetter) [reply src/pages/api/[model]/[slug]/[relatedModel].json.ts:6]: Yeah, I think you're right. I'll look into this a little more.

src/content.config.ts
```diff
@@ -1,13 +1,10 @@
...
-import { coreDataLoader } from './loaders/coreDataLoader';
-
-const models = config.detail_pages;
 
 let cols = {};
 
-if (import.meta.env.PUBLIC_STATIC_BUILD && import.meta.env.PUBLIC_STATIC_BUILD != 'false') {
-  for (const model of models) {
+if (buildStaticEndpoints) {
+  for (const model of modelTypes) {
```
ajolipa, line 7: Oops I missed this change! My bad!

src/pages/api/[model]/[slug]/[relatedModel].json.ts
```diff
@@ -1,17 +1,15 @@
...
 
 export const GET: APIRoute = async ({ params }) => {
   let data: any = {};
   const { model, slug, relatedModel } = params;
 
-  if (import.meta.env.PUBLIC_STATIC_BUILD && import.meta.env.PUBLIC_STATIC_BUILD != 'false') {
+  if (buildStaticEndpoints) {
     // @ts-ignore
     const entry = await getEntry(model, slug);
```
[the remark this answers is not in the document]
Comment (ajolipa) [reply src/pages/api/[model]/[slug]/[relatedModel].json.ts:14]: Oh my bad, somehow I missed the changes to `content.config.ts`. All good then! As long as we're okay with the build times. (But it sounds like y'all have made progress on improving that.)

src/pages/api/[model]/[slug]/[relatedModel].json.ts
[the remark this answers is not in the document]
Comment (ajolipa) [reply src/pages/api/[model]/[slug]/[relatedModel].json.ts:6]: The way it's currently set up, SSR is the default. If we want these pages built during the build step we need `prerender = true`. It should be the opposite of the flag here:
```
export const getStaticPaths = (async () => {
  if (!buildStaticEndpoints) {
    return [];
  }
...
```
(Since in the case that we're building the endpoints statically, we need to specify which paths to build; in SSR mode the parameters from the URL can be processed at request time.)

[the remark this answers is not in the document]
Comment (dleadbetter) [reply src/pages/api/[model]/[slug]/[relatedModel].json.ts:6]: Thinking a little more about this, should all of the page rendering be dependent on "output" option in `astro.config.mjs`? There are two use cases we want to serve:

1. A project is ongoing, data entry is being done, we don't want to have to rebuild the site when data is changed
2. A project is done, no more data entry will be done, we want to provide the project owner with a version of their site to be hosted on GitHub pages or Reclaim hosting

In #1, I think we'll want all of the pages to be rendered via SSR. In #2, I think we'll want all of the pages to be static. I don't think there's a use case where some are static and some are SSR, so does a single option to control how _all_ of the pages are rendered make sense?

ajolipa (approved):
This looks great to me. I guess I'd been thinking more page-by-page in terms of static vs SSR just on the theory of allowing maximal flexibility, but I agree that the real use cases are going to be all or nothing, so it makes sense to me to just set the `output` in the astro config accordingly.

src/backend/api/base.ts
jamiefolsom, line 1: @dleadbetter should API routes be versioned? Maybe internal consumers (astro templates, react  components, etc) could be expected to know which version of the API is available, but there might be external consumers of this API, who might want to opt in to newer versions on their own schedule.

dleadbetter: We certainly could version the API, would you have an example of an external consumer and how they would use the API?

jamiefolsom: @dleadbetter good question. Some sites might want to encourage external consumers by publishing endpoints in standard formats. A few that come to mind: 

- A place-centric site might want to expose their place records as a linked places gazetteer; that would be consumable by Recogito Studio's geotagging plugin.
- A media-records centric site might wants to expose its records as a iiif manifest; Faircopy or EditionCrafter might consume that manifest, or any other [IIIF client](https://canopy-iiif.github.io/docs/).
- A work/instance/item-centric site might expose its records as a bibframe bibliography. Blacklight can be made to ingest and display bibframe data, for one example.
- Pan African Data wants to make people and event data available to external scholars, maybe in some standard linked data format for events ([periodo](https://perio.do/en/)) or people (https://pelagios.org/activities/people/)

Point is **not** that we should build out those endpoints now, but rather that it would be good to be able to in the future. I realize those would almost certainly be separate endpoints and payloads from the ones you're building out now for consumption by our own internal components and templates. That might be an argument for versioning being worthwhile.

If we can build out those other endpoints/payloads for public consumption in the future, I think it makes these sites more compelling. And since those endpoints would be later additions to the static API, a version bump would make sense when we do.

Let me know what you think!

dleadbetter: @jamiefolsom - That makes sense to me, thanks for the examples. I think versioning the endpoints is definitely something we can and should do when we have external consumers. The existing endpoints only act as a pass-through to Core Data and only contain the `/api/:model/:uuid` and `/api/:model/:uuid/:related` endpoints. There are no "index" endpoints, so I'm not sure how useful they would be to an external consumer. We'd also need to do some CORS work to make sure the endpoints can be accessible across domains.

Another thing to keep in mind is that in `static` mode, the endpoints will just serve static JSON files. We won't have the ability to sort, filter, paginate, etc on demand due to the lack of a server.

PR #77, opened 2025-02-23
Adding custom primary colors and header logo

ajolipa:
### In this PR
First pass at adding the ability to customize the primary color and add a logo to the header, as well as updating the footer to the new designs and allowing logos to be added to it.

### Notes and questions
For the time being after struggling a bunch to figure out how to set Tailwind colors using asynchronously fetched data, I ended up putting the customization fields in the `settings` collection rather than the `about` collection, so that they end up in the `config.json` file that's already being copied to the `public` folder at build time. This seems to work, but if we'd prefer those fields to be in the `about` collection, then it could probably be done with an additional build script that runs before `astro build` and copies the color info to somewhere Tailwind can find it. I currently don't think there's an easy way around having to rebuild the site if the primary color is changed.

I wasn't sure exactly how much customization we need to build into the footer, so for now I kept it pretty minimal, but of course more toggles can be added for like, should the background be primary or white, or whatever.

### Screenshots

With custom (light) primary color: 
image

New category in `Settings` collection in Tina: 
image

Adding primary color: 
image

All settings: 
image

### DIFF
--- a/src/layouts/Footer.astro
+++ b/src/layouts/Footer.astro
@@ -44,7 +44,7 @@ const { title } = await fetchAbout();
               class='h-14'
               src={config.customization?.footer_logo}
             />
-            <h1 class='font-serif font-semibold text-[32px]'>
+            <h1 class='font-serif font-semibold text-[32px] text-center'>
               { title }
             </h1>
           ) : ( 
@@ -53,7 +53,7 @@ const { title } = await fetchAbout();
         }
       </div>
       <div
-        class='flex flex-col md:flex-row gap-x-12 align-baseline justify-around pt-4 pb-8'
+        class='flex flex-col md:flex-row gap-x-12 gap-y-4 items-center align-baseline justify-around pt-4 pb-8'
       >
         { _.map(_.keys(tabs), (key) => (
           <a
--- a/src/layouts/Footer.astro
+++ b/src/layouts/Footer.astro
@@ -2,90 +2,101 @@
 import { fetchAbout } from '@backend/tina-server';
 import Container from '@components/Container.astro';
 import config from '@config';
-import { ArrowTopRightOnSquareIcon } from '@heroicons/react/24/outline';
+import { Button } from '@performant-software/core-data';
+import _ from 'underscore';
 
-const { t } = Astro.props;
+const { t, tabs } = Astro.props;
 const { title } = await fetchAbout();
 ---
 
 <div
-  class='bg-white w-full'
+  class='bg-primary w-full'
 >
   <Container
-    className='flex flex-col gap-8 py-12 items-center'
+    className='flex flex-col py-12 items-center divide-y-2 divide-secondary'
   >
-    <div
-      class='flex flex-col md:flex-row gap-6 align-middle justify-center items-center'
-    >
+    {
+      config.customization?.footer_orgs && ( 
+        <div class='flex flex-row gap-12 justify-center flex-wrap py-8 w-full'>
+          { _.map(config.customization.footer_orgs, (org) => {
+            if (org.url) {
+              return (
+                <a href={org.url}>
+                  <img src={org.logo} alt={org.alt} class='h-24 object-contain' />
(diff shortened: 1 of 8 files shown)


src/layouts/Footer.astro
```diff
@@ -2,90 +2,101 @@
...
+    {
+      config.customization?.footer_login && (
+        <div
+          class='flex flex-row justify-center gap-8 py-8 w-full'
+        >
+          <a
+            href={config.core_data.url}
+            target='_blank'
+          >
+            <Button client:only rounded className='bg-white'>
```
dleadbetter, line 79: Any reason we're adding the `client:only` directive here?

src/layouts/Layout.astro
```diff
@@ -57,6 +60,7 @@ const tabs = {
   <body
     class={clsx(
       'flex h-full sm:min-h-screen flex-col bg-white text-neutral-dark mx-auto',
+      `primary-text-${primaryText}`,
```
dleadbetter, line 63: Could we use the conditional class names here rather than constructing the class name?

```
{ "primary-text-white": !useBlackText },
{ "primary-text-black: useBlackText }
```

src/utils/layout.ts
```diff
@@ -0,0 +1,24 @@
...
+        return false;
+    }
+    const r_ratio = parseInt(primary.slice(1,3), 16)/255.0;
+    const g_ratio = parseInt(primary.slice(3,5), 16)/255.0;
+    const b_ratio = parseInt(primary.slice(5,7), 16)/255.0;
+    const r = r_ratio <= 0.04045 ? r_ratio/12.92 : ((r_ratio + 0.055)/1.055)**2.4;
+    const g = g_ratio <= 0.04045 ? g_ratio/12.92 : ((g_ratio + 0.055)/1.055)**2.4;
+    const b = b_ratio <= 0.04045 ? b_ratio/12.92 : ((b_ratio + 0.055)/1.055)**2.4;
+    const l = 0.2126 * r + 0.7152 * g + 0.0722 * b;
+    if (l > 0.179) {
```
dleadbetter, line 20: Can this be simplified to `return l > 0.179`?

dleadbetter (commented):
This looks good. Store the branding in `config.json` will make it so that customizations are done by the developer during the deployment, as opposed to the end-user. Any changes to branding will require the site to be re-built. @jamiefolsom - Is this inline with the requirements?

src/layouts/Footer.astro
ajolipa: Yes, the page crashed without it. 😅 Other option would be not to use the `Button` component from `core-data`, but I felt I should use it so that any updates we make to that component apply in the footer as well for consistent aesthetics.

src/utils/layout.ts
ajolipa: Haha what a great point. 😅

src/layouts/Footer.astro
```diff
@@ -2,90 +2,101 @@
...
+              <p>
+                { t('coreDataLogin') }
+              </p>
+            </Button>
+          </a>
+          <a
+            href='/admin/index.html'
+            target='_blank'
+          >
+            <Button client:only rounded className='bg-white'>
```
dleadbetter, line 89: Will this always be `bg-white`?

ajolipa: Good question. For simplicity for now I just have the `bg-primary` version of the footer, to match the header. There could also be an inverted version with a white background and `primary` buttons, but beyond that, do we want to allow customization of the button color as well as `primary`?

ajolipa:
Update: Addressed issues above, plus a few tweaks to the mobile header to bring it in line with the new styles. I also took this opportunity to remove the max width on wide screens, because I had originally put that in by default but in fact the layout of this site is simple enough that it doesn't seem necessary. (The width limit is necessary on some more elaborate sites with background images and such to avoid weird layout spread, but here everything's just in central columns.)

Issue #69, opened 2025-02-19
Add filters counter badge to filter toggle button


dleadbetter:
Add a counter badge to the filter toggle button to display the number of filters that have been applied.

<img width="571" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOUx2Q0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--e647895917d3b415714a86b4a8ed81c7c1066db7/Screenshot%202025-02-19%20at%2011.20.44%E2%80%AFAM.png" alt="Screenshot 2025-02-19 at 11.20.44 AM.png" />

PR #83, opened 2025-02-25
CDP #69 - Filters counter

dleadbetter:
This pull request adds the filters counter to the filter toggle button.

Screenshot 2025-02-25 at 3 43 57 PM


src/apps/search/Header.tsx
```diff
@@ -36,16 +39,38 @@ const Header = (props: Props) => {
...
             onClick={() => props.onFiltersChange(!props.filters)}
             primary={props.filters}
           >
             <Icon
               name='filters'
+              size={24}
             />
+            { !_.isEmpty(items) && (
+              <div
+                className={clsx(
```
camdendotlol, line 53: This className should be a string instead of clsx.

camdendotlol (changes requested):

PR #84, opened 2025-02-26
Adding i18n functionality to `about` collection

ajolipa:
### In this PR
- Updates the schema of the `about` collection in TinaCMS to allow users to add `title`, `subheader`, and `description` for each language specified in the `config.json` file, with the default locale as top level fields and other locales as nested fields (this allows compatibility with how existing project content is configured with only one language);
- Updates the `pages/[lang]/index.astro` to fetch localized text content, and default to the default locale if no localized content is defined.

### Screenshots
image
image

### DIFF
--- a/src/i18n/utils.ts
+++ b/src/i18n/utils.ts
@@ -53,10 +53,26 @@ export const buildTranslations = (data: TranslationData) => {
   return _.defaults(translations, defaultTranslations);
 };
 
-export const getLocalizedContent = (data: any, attributes: string[], lang: string, isDefault: boolean) => {
-  let localizedContent: any = {};
-  for (const att of attributes) {
-    localizedContent[att] = isDefault || !data[lang][att] ? data[att] : data[lang][att];
+export const isEmptyRichText = (content: any) => {
+  if (!(typeof content == 'object')) {
+    return false;
   }
-  return localizedContent;
+  if (Object.keys(content).includes('children') && (!content.children || !content.children.length)) {
+    return true;
+  }
+  return false;
+}
+
+export const getLocalizedContent = (data: any, lang?: string) => {
+  if (!lang || !data[lang]) {
+    return data;
+  }
+  let localizedContent: any = {};
+  for (const att of Object.keys(data[lang])) {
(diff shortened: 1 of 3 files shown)


dleadbetter (commented):
This looks great! Would it be worth extracting the logic to a helper function so that it can be used in other places if we need to localize other content? (Possibly posts/paths in the future).

```
const { title, subheader, description } = getLocalizedContent(data, Astro.currentLocale);
```

dleadbetter:
Could you also add some screenshots of the new Tina UI?

src/i18n/utils.ts
```diff
@@ -51,4 +51,12 @@ export const buildTranslations = (data: TranslationData) => {
   });
 
   return _.defaults(translations, defaultTranslations);
+};
+
+export const getLocalizedContent = (data: any, attributes: string[], lang: string, isDefault: boolean) => {
```
dleadbetter, line 56: Is there an advantage to providing the list of attributes here?

Could we assume `isDefault` if the passed `data` doesn't contain a key for the passed `lang`?

ajolipa: Hmm, I guess I was envisioning future use cases where we're looking for different attributes, but at the cost of a bit of efficiency we can just return everything rather than specifying. (In other words with the updated version this function will return all of the data fields, not just the ones that potentially have localizations. Maybe there's a way around that but I'm somehow not seeing it -- like in this case we either need to tell the function not to return `heroImage` and `featureImage` or just go ahead and return them.)

In any case, I attempted to update it to be simpler, so see what you think.

src/pages/[lang]/index.astro
```diff
@@ -5,8 +5,17 @@ import Hero from '@components/Hero.astro';
 import { getTranslations } from '@i18n/server';
 import Layout from '@layouts/Layout.astro';
 import { TinaMarkdown } from 'tinacms/dist/rich-text';
+import config from '@config';
+import { getLocalizedContent } from '@i18n/utils';
+
+const aboutResponse = await fetchAbout();
+const { heroImage, featureImage } = aboutResponse;
+
+console.log(typeof aboutResponse.description);
```
Comment (dleadbetter) [inline src/pages/[lang]/index.astro:14]: Remove `console.log`.

dleadbetter (approved):

Issue #92, opened 2025-03-03
Bug: Clicking on the zoom controls on an embedded place in a post also opens the modal


ajolipa:
See e.g. https://gbof-c19nyc-staging.netlify.app/en/posts/Spanish-Language-Publication-Archives/#map=11.85/40.6803/-73.88309 -- zoom functionality is blocked by the modal popping up.

PR #93, opened 2025-03-03
Suppress modal when clicking on zoom controls

ajolipa:
### In this PR
Fixes the behavior described in Issue #92 by stopping propagation when clicking on the controls of an embedded map (e.g. in a post), so that the place modal does not open.


dleadbetter (approved):
Looks good 👍 . Does this also address the issue in #2? If so, let's link that issue too.

Issue #107, opened 2025-03-04
Bug: Language picker text color


ajolipa:
The text color on the language picker dropdown should be dark. 

![Image](https://github.com/user-attachments/assets/e8af7b34-9c9c-45cd-ab5d-42a74e916995)

PR #109, opened 2025-03-04
Making `LanguagePicker` dropdown text black

ajolipa:
### In this PR
Addresses Issue #107 by specifying that the text in the language picker dropdown should be black.


dleadbetter (approved):
Nice catch 👍 .

Issue #111, opened 2025-03-05
Update `TinaPlacePicker` to use Astro endpoints


dleadbetter:
Currently the `TinaPlacePicker` component is making requests directly to Core Data. We should update this to instead call the Astro endpoints.

PR #112, opened 2025-03-05
CDP #111 - Embedded Place

dleadbetter:
This pull request adds the `/api/:model/index.json` route and updates the `PlacesMap` and `TinaPlacePicker` components to use the `/backend/api/places` service to fetch data.

## Questions
In adding the `/api/:model/index.json` endpoint, I ended up copying and modifying one of the existing endpoints. It looks like the existing `/api/:model/:slug/index.json` and `/api/:model/:slug/full.json` endpoints iterate over available locales in the `getStaticPaths` functions. Is this necessary, since we don't have `/api/*` routes that are language specific?

https://github.com/performant-software/core-data-places/blob/f6af883b2275947b81f6f51061b8cf0da9f0c68b/src/pages/api/%5Bmodel%5D/%5Bslug%5D/index.json.ts#L45


ajolipa:
> In adding the `/api/:model/index.json` endpoint, I ended up copying and modifying one of the existing endpoints. It looks like the existing `/api/:model/:slug/index.json` and `/api/:model/:slug/full.json` endpoints iterate over available locales in the `getStaticPaths` functions. Is this necessary, since we don't have `/api/*` routes that are language specific?

Ah, no, good catch. Definitely a copy/paste thing that I never remembered to fix. Feel free to fix it or I can, just let me know. ^_^;

dleadbetter:
Done 👍

ajolipa (approved):
Looks good!

Issue #101, opened 2025-03-03
PAD: Facet sliders with no labels and weird values


jamiefolsom:
<img width="1640" alt="Image" src="https://github.com/user-attachments/assets/a15ba3bd-17b3-45e6-b673-a276b4ac4b5b" />

PR #113, opened 2025-03-05
Fixing display of `start_date` and `end_date` facets

ajolipa:
### In this PR
Addresses Issue #101 by adding a `date_facets` field to `config.search` that contains a list of facet names to be treated as date values _in seconds_, and updating the facet display accordingly for affected facets.
image

### Notes and questions
This feels like a bit of a brute force fix: It works for these facets, but I don't know to what extent we can assume that similar fields in other projects will also be rendered as seconds since 1/1/1970 (as opposed to milliseconds, say). Possibly a more robust solution would involve updating the `Slider` component in `react-components/core-data` to accept some sort of render prop, but then there would still be the question of where and how that render method would be configured for different facets, and still the issue of how the values display in the `currentRefinements` pills, so that seemed like possibly a much larger task to figure out.

UPDATE: The display change will now be made at the level of typesense indexing rather than by transforming the display in the facet, so this PR has now been drastically simplified just to add new i18n labels for the `start_year` and `end_year` facets. ^_^;


dleadbetter (commented):
Would it be more straight-forward to update the way we're indexing these records into Typesense? We could adjust these values to be indexed as numeric years instead of Unix timestamp values.

There's also a question around whether it makes sense for the `start_date` and `end_date` to be two separate facets?

ajolipa:
> Would it be more straight-forward to update the way we're indexing these records into Typesense? We could adjust these values to be indexed as numeric years instead of Unix timestamp values.
> 
> There's also a question around whether it makes sense for the `start_date` and `end_date` to be two separate facets?

All great questions! I guess I defaulted to transforming it on the front end just because I know that code and don't really know the typesense indexing code, but that is not a real reason. 😅 Are there use cases where we would want events indexed in Typesense at a more granular level than year? If so where would that be determined?

Re: a combined facet, that would certainly make sense, but seems like it would need a moderate amount of additional configuration and a new component that can accept multiple attributes. 🤔 Certainly doable if we think it's worth it.

PR #123, opened 2025-03-12
Adding basic Core Data media embeds to Tina RTE

ajolipa:
### In this PR
Adds a Core Data Media embed to the RTE of paths and posts, which loads data from the `media_contents` endpoint of the public API and allows the user to select any media record from the associated core data projects. 
image

On the front end, the media is displayed using the Clover IIIF viewer. 
image

### Notes and questions
- What sort of flexibility do we want to provide in terms of how the viewer is displayed? I figure the point of getting the data from CD rather than just uploading an image is largely to use the IIIF viewer so I'm just going with that for now, but if we want any more configuration or styling customization I can work on building that in.
- This functionality depends on the `/media_contents` endpoint of the public core data API existing, which it might not quite yet (it's been created but maybe pending the next release?).


src/apps/paths/PathViewer.tsx
```diff
@@ -17,6 +17,7 @@ import {
   useState
 } from 'react';
 import { TinaMarkdown } from 'tinacms/dist/rich-text';
+import CoreDataMedia from '@components/CoreDataMedia';
```
dleadbetter, line 20: Could we rename this something like `Media` or `MediaInsert` to be consistent with the `Place` component?

src/components/CoreDataMedia.tsx
```diff
@@ -0,0 +1,13 @@
+import Viewer from "@samvera/clover-iiif/viewer"
```
dleadbetter, line 1: Code style seems a little off in this file: indentation and double quotes

src/content.config.ts
```diff
@@ -11,7 +11,7 @@ if (isStaticBuild) {
     cols[model] = defineCollection({
       loader: coreDataLoader({
         model: model,
-        getRelations: true
+        getRelations: model !== 'media_contents'
```
dleadbetter, line 14: Hm, I see what we're trying to do here. We don't want to allow related endpoints for `media_contents` records. Is there a better way to handle this than having `model !== 'media_contents'` checks in a bunch of spots?

This could also come back to the question of whether we want to load related records as separate API calls.

tina/components/TinaMediaPicker.tsx
```diff
@@ -0,0 +1,165 @@
...
+  const [selectedMedia, setSelectedMedia] = useState<any>();
+  const [message, setMessage] = useState('');
+  const [loading, setLoading] = useState(false);
+
+  function classNames(...classes) {
+    return classes.filter(Boolean).join(' ')
+  }
+
+  const fetchMedia = useCallback((params = {}) => {
+    const url = getMediaURL(params);
```
dleadbetter, line 27: This looks to be making a request directly to the Core Data API. Should we instead be making the request to the Astro API?

tina/components/TinaMediaPicker.tsx
```diff
@@ -0,0 +1,165 @@
...
+      content_url: e.content_url,
+      content_preview_url: e.content_preview_url
+    };
+    props.input.onChange(newData);
+  };
+
+  useEffect(() => {
+    setLoading(true);
+    MediaContentsService
+      .fetchAll({ per_page: 0 })
```
dleadbetter, line 46: Could this get problematic if a project has thousands of media records?

tina/components/TinaMediaPicker.tsx
ajolipa: Oops, meant to update that, good catch!

src/content.config.ts
ajolipa: Yyyeah this isn't the most elegant. Could change the list of model types to have a flag on each one specifying which endpoints should be generated for it, but that also felt like overkill when there's just one model type that's excluded. Open to suggestions.

tina/components/TinaMediaPicker.tsx
dleadbetter: If the Place selection component is also doing this, I think it could also cause performance issues. After thinking about it some, I don't have a good solution. Paginate will work if we're loading records from the Core Data API, but not if we're statically rendering the site and loading records from Astro's content layer. I guess for now we can leave this as is and solve the performance issue when we get there.

Another question on the Place selection component: Is this component similar enough to the Place selector that we could extract some of the logic to a component used by both the Place picker and Media picker?

tina/components/TinaMediaPicker.tsx
ajolipa: Actually I think with the updates to the place picker and media picker to get data from the Astro endpoints, I think the whole `tina/utils` file is not needed anymore.

tina/components/TinaMediaPicker.tsx
ajolipa: Good question, there's definitely some overlap, but the fields are not the same and the place picker has a bunch of extra config options... I'll take a look and see if there's logic about fetching the data that could be extracted out though.

dleadbetter: I was thinking particularly about the JSX, then components used for selecting look to be similar.

ajolipa: Yeah, has been updated now.

src/apps/posts/PostContent.tsx
```diff
@@ -28,7 +29,8 @@ const PostContent = (props: PostContentProps) => (
         <TinaMarkdown
           components={{
             place: PlaceInsert,
-            iframe: IframeEmbed
+            iframe: IframeEmbed,
+            core_data_media: MediaInsert
```
dleadbetter, line 33: Can we change the key to just `media`?

tina/components/TinaModelPicker.tsx
```diff
@@ -0,0 +1,135 @@
+import { Combobox } from '@headlessui/react';
```
dleadbetter, line 1: Nice! 👍

src/apps/posts/PostContent.tsx
ajolipa: Sure; I guess I wanted to be specific since we might have other embedded components in the future that allow media from other external sources, whereas this component is only drawing from the data in core data, but I suppose we can disambiguate that if necessary in the future.

dleadbetter (approved):

Issue #128, opened 2025-03-14
Add option for using native TinaCMS user management


ajolipa:
On a site-by-site basis it should be possible to opt out of Keycloak SSO entirely and use the native TinaCMS user management system instead. I believe that implementing this would involve the following steps:

- In `/tina/config`, modify [this line](https://github.com/performant-software/core-data-places/blob/8bd399c04947578eabc79bfb5e362a057ae828a1/tina/config.ts#L17) to look for the presence of the `AUTH_KEYCLOAK_ISSUER` env variable and if not found use the `UsernamePasswordAuthJSProvider()` instead of the custom one.
- Include the `users` collection in the Tina schema when relevant.
- Update the `tina` netlify function to again check for the existence of a Keycloak-related env variable, and if not found use the default `AuthJsBackendAuthProvider` [here](https://github.com/performant-software/core-data-places/blob/8bd399c04947578eabc79bfb5e362a057ae828a1/netlify/functions/tina.ts#L65). 

I believe those three places are all that would have to change, plus adding the `tinacms-authjs` dependency. Then if a project is not configured to use Keycloak it will default to managing users inside Tina.

PR #129, opened 2025-03-14
Adding non-Keycloak fallback user management

ajolipa:
### In this PR
Initial attempt at addressing Issue #128 by defaulting to use Tina's native user management system if no `AUTH_KEYCLOAK_ISSUER` is configured in the environment variables.


ajolipa:
> Looks good! Just a couple comments on very minor issues. Do we also need to include a JSON file in order to seed the user collection?

That would be in the various content repos, wouldn't it? Or should it also be here somewhere?

I will fix the nits and re-submit for review momentarily. My other question is just how worried I should be about the PSS scavenger hunt deploy preview not working, given that the others do. It seems like it's hitting an error while trying to install `better-sqlite3`? but I'm not sure where that's coming from or why the other sites aren't hitting the same error. :-/

dleadbetter:
> That would be in the various content repos, wouldn't it? Or should it also be here somewhere?

Ah right. There is a section in the `README.md` about how to obtain that seed data: https://github.com/performant-software/core-data-places/blob/develop/README.md#users

Should we add back the sample file here so it can be easily copied into the content repos?

dleadbetter:
> My other question is just how worried I should be about the PSS scavenger hunt deploy preview not working, given that the others do.

Ideally they should all build. It's a little curious that one site is not working when they are all deploying from the same code base. Could we check that all sites are using the same versions of Node and any other configurable dependency to see if we can pinpoint where the discrepancy is?

ajolipa:
> > My other question is just how worried I should be about the PSS scavenger hunt deploy preview not working, given that the others do.
> 
> Ideally they should all build. It's a little curious that one site is not working when they are all deploying from the same code base. Could we check that all sites are using the same versions of Node and any other configurable dependency to see if we can pinpoint where the discrepancy is?

Aha, the scavenger hunt site was for some reason using a different build image (Ubuntu 20.04 rather than 24.04). Updated that now, hopefully will fix the issue.

ajolipa:
I did also add back the `/data/users.json` sample file and udpated the README to mention that that step should be skipped if you're using SSO, so ready for re-review whenever I think @dleadbetter .

.env.example
```diff
@@ -47,4 +47,13 @@ S3_FOLDER=subfolder1
...
+TINA_PUBLIC_MAP_LIBRARY_URL=https://...
+
+# Configuration for Keycloak SSO login if relevant
+AUTH_KEYCLOAK_ID=tinacms
+AUTH_KEYCLOAK_ISSUER=https://keycloak.archivengine.com/realms/your-realm
+AUTH_KEYCLOAK_SECRET=123
+AUTH_SECRET=123
+AUTH_TRUST_HOST=true
+PUBLIC_BASE_URL=http://localhost:8888/
+TINA_PUBLIC_AUTH_USE_KEYCLOAK=true
```
dleadbetter, line 59: Sorry, should have seen this earlier. Let's change the default for `TINA_PUBLIC_AUTH_USE_KEYCLOAK` to "false" so that copying `.env.example` doesn't result in an error if that is not changed.

ajolipa: Oh yes fantastic point. Done.

dleadbetter (approved):

Issue #131, opened 2025-03-17
Related records not displaying on detail panel


dleadbetter:
When viewing a record detail panel, only the related place records are displaying.

<img width="1208" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNGRKQ1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--79ffc41beb1c16e80999bd41da4e7774c08edc60/Screenshot%202025-03-17%20at%2011.06.52%E2%80%AFAM.png" alt="Screenshot 2025-03-17 at 11.06.52 AM.png" />

PR #132, opened 2025-03-17
Bug fix: Added `relatedPeople` back to record detail panel

ajolipa:
### In this PR
Addresses at least part of Issue #131 by adding a line that was missing to the `relations()` function in `BasePanel`, so that related people records are properly included.


dleadbetter (commented):
Looks good to me! Is there more to fix to fully close out the issue? The PR description specifies "at least part".

ajolipa:
> Looks good to me! Is there more to fix to fully close out the issue? The PR description specifies "at least part".

That was just because at the time it wasn't clear that the observed problem was only with related people, but I believe there were not in fact examples with other model types, so we should be good.

PR #138, opened 2025-03-18
Develop into master

dleadbetter:
Merging `develop` into `master` for a production release.


ajolipa (approved):
Not going individually through all 136 files here but as best I can tell everything's working nicely on the deploy previews. As long as all the existing GBoF sites that are deploying from `master` have auto published turned off so we can migrate them individually, this seems good to go as far as I'm concerned!

dleadbetter:
Yes, auto-publishing is turned off for every site that deploys from `master`. The deploy previews will still build (and some likely fail), but it will not auto publish.

Issue #26, opened 2025-01-10
Add linting/formatting configuration


dleadbetter:
To keep code style consistent, we should add a linting/formatting configuration.

camdendotlol:
This might be a good opportunity to continue with the standardized linting config Nick and I were trying out in EditionCrafter: https://github.com/antfu/eslint-config with the included VSCode `settings.json` for autoformatting.

PR #143, opened 2025-03-20
Model detail pages

camdendotlol:
# Summary

- adds a new `RecordDetail` Astro component that shows a specific record on the `[lang]/[model]/[slug]` route
- includes special handling for model-specific fields, such as showing places on a map and showing start/end dates for events
- uses the existing `UserDefinedFieldView` component for user-defined fields

## Screenshot

(zoomed out so I could fit the whole page in)

Screenshot 2025-03-20 at 4 53 08 PM


src/apps/search/panels/BasePanel.tsx
```diff
@@ -155,6 +166,13 @@ const BasePanel = (props: Props) => {
   },
   [data, props.name]);
 
+  /**
+   * Gets the detail page URL for the passed item.
+   */
+  const getItemUrl = useCallback((model, data) => {
+    return `/${lang}/${model}/${data?.uuid}`
+  }, [lang])
```
dleadbetter, line 174: Missing semi-colons.

src/apps/search/panels/BasePanel.tsx
```diff
@@ -336,6 +354,9 @@ const BasePanel = (props: Props) => {
         onClose={() => navigate('/')}
         relations={relations()}
         title={name}
+        detailPageUrl={config.detail_pages.includes(plurals[props.name])
```
dleadbetter, line 357: Would it make sense to do this logic in the model-specific implementations of the component and pass the URL as a prop?

src/components/RecordDetail/Relations.astro
```diff
@@ -0,0 +1,49 @@
+---
```
dleadbetter, line 1: We have an existing `AccordionItemsList` component that might be useful here.

src/components/RecordDetail/PlaceMap.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+interface Props {
+  record: any;
+}
+
+const PlaceMap = (props: Props) => {
+  return (
+    <Peripleo>
+      <Map
+        style={style}
```
dleadbetter, line 14: Is there any reason we're not using the map styles specified in `config.json`?

src/pages/[lang]/[model]/[slug]/index.astro
```diff
@@ -0,0 +1,56 @@
...
+import config from "@config";
+import RecordDetail from "@components/RecordDetail/index.astro";
+import { getRelations } from "@loaders/api/helpers";
+import { relatedModelTypes } from "@loaders/coreDataLoader";
+
+const { model, slug } = Astro.params;
+
+let record: any;
+
+if (import.meta.env.STATIC_BUILD) {
```
Comment (dleadbetter) [inline src/pages/[lang]/[model]/[slug]/index.astro:13]: This Astro API acts as an intermediary and should decide whether to fetch the data from Core Data or the content layer. There are service classes in `/backend/api` that can be used to call the Astro API.

dleadbetter (commented):
Looking good! A couple of comments and a note on the code organization: 

I think using `[model]` as a parameter could get messy. It requires a bunch of `model === this ? doSomething() : doSomethingElse()` logic across different components and maintain maps of model names.

Could we instead structure the pages as: `/pages/[lang]/events/[slug]/index.astro`? This way we could have model-specific implementations and also use a common component to do the heavy lifting without the need for the common component to ever need to know what type of model is being rendered. For example:

```
// /pages/[lang]/events/[slug]/index.astro

<RecordDetail
  renderItem={(event) => (
    <div>
      { renderStartDate(event) }
      { renderEndDate(event) }
    </div>
  )}
  service={EventsService}
/>
```

Then the common component has all the information it needs to fetch the record, fetch related records, render user-defined fields, etc. Check out the detail side panel components in `/src/apps/search/panels` for an example.

Also, on code style, let's double check for the following:
- Single quotes
- Using new lines in JSX after props
- No ternary operators in JSX unless necessary
- Consistent indentation
- Use of semi-colons

src/components/RecordDetail/PlaceMap.tsx
camdendotlol: Which config object contains map styles? I see `layers`, but I don't see `version` or `sources` in the Scavenger Hunt config file.

src/pages/[lang]/[model]/[slug]/index.astro
Comment (camdendotlol) [reply src/pages/[lang]/[model]/[slug]/index.astro:13]: Is there a way to get all related records without calling `fetchRelatedEvents`, `fetchRelatedInstances`, `fetchRelatedItems`, etc. sequentially?

src/components/RecordDetail/PlaceMap.tsx
dleadbetter: You should be able to use [PeripleoUtils](https://github.com/performant-software/react-components/blob/master/packages/core-data/src/utils/Peripleo.js#L58) to convert the layers defined in the config. See [MapView](https://github.com/performant-software/core-data-places/blob/develop/src/apps/search/MapView.tsx) for an example.

src/pages/[lang]/[model]/[slug]/index.astro
Comment (dleadbetter) [reply src/pages/[lang]/[model]/[slug]/index.astro:13]: Currently no, there isn't, but something we could look to add. Is there an advantage to getting all of the related records in a single API call vs. using multiple?

dleadbetter:
One other thing I just thought of: We'll probably want the same ability to suppress fields/relationships on the detail pages that exists on the detail side panels using the `search.result_filtering` property.

src/pages/[lang]/[model]/[slug]/index.astro
Comment (camdendotlol) [reply src/pages/[lang]/[model]/[slug]/index.astro:13]: Am I correct in my understand that the Astro API routes are sort of the official clearing house for records? i.e. anything other than the search (which gets data from TS) should get data via the Astro API routes?

One thing I'm noticing is that it's difficult to call the API routes from the Node side (SSR-ing a page or when building in SSG mode). Node doesn't seem to like the relative URL passed to `fetch` in https://github.com/performant-software/core-data-places/blob/develop/src/backend/api/base.ts#L145. I assume it would work with a domain prepended but it still feels un-idiomatic to have the Astro server making requests to its own API routes, and I'm still not sure that would fix the issue in SSG mode.

So if the idea is that the API routes are the primary source of data, we might need to factor out all their logic into helper functions that can be shared between the API routes and server-side components.

Actually, just for clarity, in which scenarios do we need to call the CD API as opposed to the content collection? Is there a reason the content collection can't take care of the situations in which we're calling CD directly?

Comment (dleadbetter) [reply src/pages/[lang]/[model]/[slug]/index.astro:13]: That's correct. The Astro API should be used across the application to fetch records from Core Data. You're right, the classes in `/backend/api` are only going to be able to be used on the client. I think we'll likely need to do something like [this](https://docs.astro.build/en/recipes/call-endpoints/) in the Astro components to call the API on the server side.

The Core Data API will be called directly when we're running in `server` mode (e.g when `STATIC_BUILD` is "false"). The content collections are populated at built time by the loaders when we're running in `static` mode. We can fetch records from the content collections, but they'll be empty when we're running in `server` mode. The Astro endpoints are designed to handle the business logic for where to fetch the data: In `static` mode, we'll get the data from the content collections, in `server` mode we'll get the data from Core Data. This way, the individual components don't need to be aware of how we're building the site, they just always get data from the same place.

Comment (dleadbetter) [reply src/pages/[lang]/[model]/[slug]/index.astro:18]: Let's move forward with extracting the data fetching logic out of the API and into a common location that can be used by both the API and the SSR pages.

Screenshot 2025-03-24 at 12 10 59 PM

src/components/RecordDetail/Relations.astro
camdendotlol: I don't think the accordion would match the [designs](https://www.figma.com/design/5AWa1c3pJeVXwEeVPbRqZj/Place-Components-2.0?node-id=713-4649&m=dev). I'm also not sure that being able to collapse/expand the lists of relations would be desired on a detail page.

src/apps/search/panels/BasePanel.tsx
camdendotlol: What if I open a separate PR to add ESLint and some sort of basic linting config? It seems time-consuming to go through the files and manually replace quotation marks, etc. I see there's an issue for it already: https://github.com/performant-software/core-data-places/issues/26

src/components/RecordDetail/MediaContents.tsx
```diff
@@ -0,0 +1,37 @@
+import { useState } from 'react';
+import { useTranslations } from '@i18n/client';
+import { MediaGallery } from '@performant-software/core-data';
+
+interface Props {
+  data: any[];
+  uuid: string;
+}
+
+const MediaContents = (props: Props) => {
```
dleadbetter, line 10: Here we should be displaying collections of manifests, much like we do in the detail panels.

src/components/RecordDetail/UserDefined.astro
```diff
@@ -0,0 +1,29 @@
...
+const { record } = Astro.props
+
+const lang = Astro.currentLocale;
+const { t } = await getTranslations(lang);
+---
+{record.user_defined && (
+  <>
+    {Object.keys(record.user_defined).map(uuid => (
+      <>
+        {config.detail_pages?.exclude_fields && !config.detail_pages?.exclude_fields.includes(uuid) && (
```
dleadbetter, line 15: The exclude fields should also be able to suppress related records and native fields. See the implementation in the detail side panels.

Is there any reason we wouldn't want to use the same config option for the side panels and detail pages?

dleadbetter:
A couple of other notes testing locally:

### I'm not longer able to see any related records on the detail panels
Screenshot 2025-03-25 at 4 17 00 PM

### Detail pages appear empty
Screenshot 2025-03-25 at 4 17 10 PM

src/components/RecordDetail/MediaContents.tsx
```diff
@@ -0,0 +1,37 @@
+import { useState } from 'react';
```
dleadbetter, line 1: Would it make sense to move these components from `components/*` to something like `apps/pages/*`? The idea of the `apps` directory is to isolate components that are specific to certain parts of the app, but not necessarily used anywhere else.

src/components/RecordDetail/PlaceMap.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+interface Props {
+  record: any;
+}
+
+const PlaceMap = (props: Props) => {
+  const { baseLayers } = PeripleoUtils.filterLayers(config);
+
+  return (
+    <Peripleo>
```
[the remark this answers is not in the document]
camdendotlol: The `Map` component errors out when trying to read the config from `useRuntimeConfig`. I assume `useRuntimeConfig` has something to do with search state that's not available when running in the detail page?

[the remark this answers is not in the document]
dleadbetter: I think you should be able to wrap it in a `<RuntimeConfig>` component, much like `PostContent` is doing.

src/components/RecordDetail/index.astro
```diff
@@ -0,0 +1,80 @@
...
+    <h1 class="font-bold text-3xl text-neutral-800 mt-0 mb-6">{record.name || getNameView(record)}</h1>
+    {record.description ? <p class="text-neutral-800 py-6 break-words">{record.description}</p> : null}
+    {record.biography ? <p class="text-neutral-800 py-6 break-words">{record.biography}</p> : null}
+    <slot />
+    <UserDefined record={record} />
+    {Object.keys(relations).map((uuid) =>
+      {
+        const { data, model } = relations[uuid]
+
+        if (model === 'media_contents') {
```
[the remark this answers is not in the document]
camdendotlol: > Also, we should be using manifests here, not the individual media records.

Can you expand on this? I'm not really sure what you mean. The `MediaGallery` model opens with a IIIF viewer when you click on a thumbnail.

src/components/RecordDetail/UserDefined.astro
camdendotlol: > Is there any reason we wouldn't want to use the same config option for the side panels and detail pages?

Mixed feelings about this. If the user wants to filter the same set of fields it would be inconvenient to have to enter the data twice. But I can imagine them wanting to filter a different set of fields in each place.

camdendotlol:
@dleadbetter 

For your second note about detail pages being empty, I changed the config syntax for `detail_pages` so it looks like this:

```
  "detail_pages": {
    "models": [
      "events",
      "organizations",
      "items",
      "places",
      "people",
      "works"
    ],
    "exclude_fields": {
      "events": ["6915e33c-a864-41e2-a1c3-7d54e4deb984"],
      "people": ["biography"],
      "places": ["60fbb406-3a7c-4459-a1df-699c23f23c82"]
    }
  },
```

src/components/RecordDetail/index.astro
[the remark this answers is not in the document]
dleadbetter: We should use the `/public/v1/<model>/<uuid>/manifests` endpoint to get a IIIF collection manifest. Each collection represents a relationship from the base record to a set of media records. The items displayed under "Media" will be the thumbnail from the collection. Clicking on the thumbnail should open the IIIF viewer for a single manifest, which contains all the images in the relationship. See the detail side panels for implementation details.

src/components/RecordDetail/UserDefined.astro
dleadbetter: Let's discuss this in our meeting today.

src/components/RecordDetail/index.astro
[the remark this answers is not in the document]
camdendotlol: I noticed the panels make a separate request to the `/api/[model]/[uuid]/[relatedModel].json` endpoint to fetch the manifest because the `full.json` endpoint doesn't include manifests, event though it includes all other relations. Would it make sense to include related manifests in `getRelatedRecords`?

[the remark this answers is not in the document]
dleadbetter: Yes, in fact, I think it would make sense for `getRelatedRecords` to include `manifests` and _not_ include `media_contents`. We've kind of standardized displaying related media as IIIF manifests, so I don't believe there's anywhere we're actually calling `/api/<model>/<uuid>/media_contents`. As a result, the data loader is likely doing unnecessary work to fetch that data.

camdendotlol:
@dleadbetter Since we're no longer using a separate array for excluded fields on the detail page, I've reverted the `detail_pages` config syntax to the original, e.g.

```js
"detail_pages": ["items", "people"]
```

src/apps/pages/RecordDetail/PlaceMap.tsx
```diff
@@ -0,0 +1,43 @@
...
+        path='/config.json'
+        preprocess={PeripleoUtils.normalize}
+      >
+        <Peripleo>
+          <Map>
+            <LocationMarkers
+              boundingBoxOptions={{
+                animate: false
+              }}
+              data={props.record.place_geometry.geometry_json}
```
dleadbetter, line 34: This should probably have some `null` checking. Is it possible to get to this page for a place without a geometry?

netlify.toml
```diff
@@ -16,6 +16,7 @@
 [dev]
   command = "npm start"
   targetPort = 4321
+  autoLaunch = false
```
dleadbetter, line 19: Any reason we're disabling this?

src/apps/pages/RecordDetail/PlaceMap.tsx
camdendotlol: There's a null check before the return.

dleadbetter (approved):

Issue #146, opened 2025-03-24
Make table results view optional


ajolipa:
Not all projects want to include the table view for search results. It should be possible to configure whether or not that view is included as an option in the config.

PR #147, opened 2025-03-24
Makes table view for search results opt-in

ajolipa:
### In this PR
Addresses Issue #146 by adding a `search.table` flag to the `config.json` file that opts in to including the table view in the search results interface, since currently the majority of projects don't actually want that view available.

### Questions
Should it be opt-out instead? I did it this way just because it will involve editing fewer content repos right now, but I could definitely see doing it the other way; open to suggestions.


src/apps/search/SearchLayout.tsx
```diff
@@ -31,6 +31,8 @@ const SearchLayout = () => {
   const route = useCurrentRoute();
   const id = getCurrentId(route);
 
+  console.log(config);
```
dleadbetter, line 34: Remove console.log

dleadbetter (commented):
Looks good. For most config options, we give the user everything and allow them to suppress things they don't want (facets, fields on the side panels, etc). Would it make sense for this config to also follow that convention and be opt-out?

absempere:
@dleadbetter @ajolipa Yes, I think it makes sense to stick with the established conventions and make this opt-out.

ajolipa:
> @dleadbetter @ajolipa Yes, I think it makes sense to stick with the established conventions and make this opt-out.

Okay, updated it so that it's opt-out. Since we want it suppressed for GBoF sites this will involve updating all of those config files accordingly whenever this change is released.

src/apps/search/SearchLayout.tsx
```diff
@@ -94,6 +94,7 @@ const SearchLayout = () => {
         onFiltersChange={setFilters}
         onViewChange={setView}
         view={view}
+        tableView={!(config.search?.table === false)}
```
dleadbetter, line 97: Will `config.search.table` be undefined if not set? Can we just pass `config.search.table` here and default it to "false" in `Header`?

README.md
```diff
@@ -32,7 +32,6 @@ Copy the `/data/users.json` file into your content repository to `/content/users
 | Key                                           | Type    | Description                                                                                                                                                              |
 |-----------------------------------------------|---------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | content                                       | Array   | Array of content keys to allow for data entry and routing: "posts", "paths"                                                                                              |
-| detail_pages                                  | Array   | Models to be shown on detail pages.                                                                                                                                      |
```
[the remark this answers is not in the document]
ajolipa: Oops, no, that must have come from a merge after I first made the branch and I didn't rebase correctly somehow. Will fix.

src/apps/search/SearchLayout.tsx
ajolipa: Ah yeah that is more elegant way to turn `undefined` into `true`. 😅

dleadbetter (approved):

Issue #142, opened 2025-03-20
Multiple search indexes per site


dleadbetter:

dleadbetter:
## Testing Notes
The https://pss-scavenger-hunt.netlify.app/ staging site has been updated to display two separate search indexes: places and events. There's not much specific to test other than both searches work as expected. General smoke testing would be good too, as this change touched a lot of areas on the site.

PR #149, opened 2025-03-26
CDP #142 - Multisearch

dleadbetter:
This pull request implements the ability to include multiple search indexes in a single site. Each view will be backed by a different Typesense index, use different configuration options, and include different facets.

Screenshot 2025-03-27 at 10 49 02 AM

Screenshot 2025-03-27 at 10 50 22 AM

As a result, the `search` config object has been converted to an array to allow configuring multiple indexes separately.

Screenshot 2025-03-27 at 10 52 02 AM

Screenshot 2025-03-27 at 10 52 06 AM


build.config.mjs
```diff
@@ -14,11 +14,12 @@ const init = async () => {
     console.info('Using remote config.json');
   } else if (fs.existsSync('./public/config.dev.json')) {
     fs.copyFileSync('./public/config.dev.json', './public/config.json');
-
     console.info('Copying config.dev.json');
   } else {
     console.info('Using local config.json');
   }
+
+  console.info('');
```
blms, line 22: Stray debug log?

blms (approved):
This looks good! Surprising how many changes are required just due to changing a config object to an array 😆 but makes sense given how the whole site is built from the config. I like the approach of using a `SearchContext` that extracts the config using the current search name.

No issues jumped out at me, but I think this one will definitely benefit from a thorough smoke test, since it touches pretty much every part of the application.

Issue #151, opened 2025-03-28
Disable TinaCMS on static deployment


dleadbetter:
When deploying statically TinaCMS should not be available.

PR #156, opened 2025-04-01
CDP #151 - Disable TinaCMS

dleadbetter:
This pull request updates the `Footer` component to hide the Core Data and TinaCMS login buttons when we're building the site in "static" mode. 

It also adds the optional `USE_CONTENT_CACHE` environment variable to allow the site to be built in "static" mode without running the content loaders. This is useful in a dev environment when the dev server may be restarted multiple times.

Screenshot 2025-04-01 at 7 26 47 AM

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -49,6 +49,10 @@ STATIC_BUILD=false
 # URL for a georeferenced map library JSON, which should retrieve an array of objects with "name" and "url" keys.
 TINA_PUBLIC_MAP_LIBRARY_URL=https://...
 
+# If true, content loaders will not be run. This option is useful in development environments when we don't want to
+# load all of the data each time the development server restarts.
+USE_CONTENT_CACHE=false
+
 # Configuration for Keycloak SSO login if relevant
 AUTH_KEYCLOAK_ID=tinacms
 AUTH_KEYCLOAK_ISSUER=https://keycloak.archivengine.com/realms/your-realm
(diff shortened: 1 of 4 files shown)


dleadbetter:
@ajolipa - Does this approach make sense? I know you mentioned writing content loaders for Tina. I've found that even when statically generated, the GraphQL endpoints for Tina are still available, but since there's no edge function, there's no way to authenticate. I figured we could just start up Tina, use the GraphQL endpoints to fetch the data at build time, then hide the login button.

ajolipa:
> @ajolipa - Does this approach make sense? I know you mentioned writing content loaders for Tina. I've found that even when statically generated, the GraphQL endpoints for Tina are still available, but since there's no edge function, there's no way to authenticate. I figured we could just start up Tina, use the GraphQL endpoints to fetch the data at build time, then hide the login button.

Yeah, this sounds morally equivalent to a data loader. I guess my only question is whether this depends on the MongoDB backend for Tina existing at build time; I'm picturing a case like, the client statically generates the site, then wants to make a small edit to some text or whatever but doesn't want to re-subscribe to our Tina instance -- in theory they could just edit the content in the content repo directly, then rebuild the site with the new content. If the MongoDB that was behind their Tina instance was shut down, would that rebuild work? (Maybe the answer is yes, or maybe this is such a corner case that it doesn't matter lol, but that's my only question/reservation about using the GraphQL endpoint rather than basically pulling content from the github repo directly into the Astro data layer.)

Issue #116, opened 2025-03-06
Create timeline view


dleadbetter:
## Overview
Create a timeline view in the search interface. Events displayed in the timeline should come from the events related to the Typesense search results. This will require a new configuration option to provide the relationship between the base record and the related events. If the `path` attribute is empty, we should assume that the records indexed are events themselves. 

```
"search": {
   "timeline": {
    "visible": true,
    "path": "7df1f742-a69a-45db-ad97-1f405e3ba99f"
  }
}
```

## Designs
https://www.figma.com/design/5AWa1c3pJeVXwEeVPbRqZj/Place-Components-2.0?node-id=15-15372&t=ZpjCk5bbpkm4wq0Y-1

jamiefolsom:
I would add a couple of points, to explain my understanding of aspects of the design.

1. For reference, the design is based on TimelineJS from the KnightLab. I am not against [using that](https://github.com/NUKnightLab/TimelineJS3?tab=readme-ov-file#use-via-es6-moduleswebpack) if it's a good solution. If it is not, of course, not is fine too.
2. Events should be stacked vertically so as to make as many of them legible as possible. When they're bunched up on a given date, mouseover should bring the hovered event to the front.
3. Clicking the event should load it in the detail panel, and ideally, if there's a related place or places, zoom there and highlight that place (or places) on the map, and select any related search result(s) in the list or table.
4. It would be good to be able to have anchors for events that are represented either as points on the line (as in the design), or when they have start/end dates, as ranges/bars on or above the line.
5. I _think_ The zoom buttons should change the scale of the timeline, not change the position of the slider as is currently the case. But I could be convinced otherwise. Might want to ask @cblissg. When the timeline is zoomed "all the way out", which I define as "the smallest timeline possible while showing all events", zooming in would make the timeline ticks for dates further apart. Zooming back out would 
6. Only visible events should be shown in search results, map, etc. It's a facet, in other words.

blms:
A note and question from my work so far:
- TimelineJS unfortunately is not customizable enough for our needs (no callback on clicking events, not possible to integrate with facets, no customization of zoom behavior, requires use of "slides"). This is fine as our existing component is totally workable.
- Am I understanding correctly that we actually don't want a slider with two endpoints here, and the facet should instead be controlled entirely by timeline zoom/pan? I can ask Chelsea about this as well if we're not sure.

PR #158, opened 2025-04-01
Timeline (#116)

blms:
## In this PR

Per #116:
- Adds configuration to `config.json`/Tina for the timeline. The values are:
  - `date_range_facet` - Path to the facet field in Typesense that the timeline will interact with. This is usually something like `"event_range_facet"` or `"start_year_facet"`. This is required for the Timeline button to appear on a search page.
  - `event_path` - Path to the event relations list for the primary documents in Typesense. This is required if Event is not the primary model for the search index, and required to be unset if Event is the primary model.
- Shows the timeline toggle button, and the timeline itself if clicked, on the search page. Disables the "Table" button if the timeline is open, and vice versa.
- For display on the timeline: Pulls a list of related events from current search hits if Event is not the primary model, and otherwise just uses the whole list of hits.
   - On click, navigates to the event using the `/events` route.


dleadbetter:
Looks good! Could you add some screenshots? I'm also curious about explicitly settings the `width` on the `TimelineView`. Maybe some screenshots will help me understand why that's necessary.

blms:
Sure thing. The reason the width needs to be explicitly set is that the `FacetTimeline` component is set to fill horizontally whatever container it's inside, and scroll its overflow when zoomed. As a result, the width of its container must be explicitly set, in one way or another. Setting it to some percent value or other "fill" kinds of width values always resulted in it overflowing when different panels were opened, or when the zoom buttons were pressed. Maybe there is a better way to do this with CSS that involves modifying the `FacetTimeline` component!

No panels opened, zoomed out and zoomed in:

Screenshot 2025-04-01 at 1 41 57 PM
Screenshot 2025-04-01 at 1 42 55 PM

Both filter and detail panels opened:

Screenshot 2025-04-01 at 1 37 54 PM

dleadbetter (approved):
Very nice, this looks great 👍 .

PR #160, opened 2025-04-01
Troubleshooting image uploads

ajolipa:
### In this PR
Attempting to fix media uploads via Tina. Adding the catch-all route for `/api/s3/images/*` seems to fix one level of the issue, but now I am running into a different issue where it seems that the `S3_FOLDER` environment variable isn't being used correctly and it's just trying to upload the image directly to the parent folder of the bucket. 
image
image

### DIFF
--- a/netlify/functions/tina.ts
+++ b/netlify/functions/tina.ts
@@ -107,6 +107,8 @@ app.get('/api/tina/*', async (req, res) => {
 
 app.get('/api/s3/media', mediaHandler);
 
+app.get('/api/s3/media/*', mediaHandler);
+
 app.post('/api/s3/media', mediaHandler);
 
 app.delete('/api/s3/media/:media', (req, res) => {


dleadbetter:
I think the new issue occurs [here](https://github.com/tinacms/tinacms/blob/95a293cffab73635dc0677ad5277fc84cbf2c507/packages/next-tinacms-s3/src/handlers.ts#L82). When I originally implemented this, we were using v1.0.0 of the `next-tinacms-s3` package (eek 😟 ). Since then, they've added uploads via signed URLs, but it doesn't look like the signed URL request takes into consideration the `mediaRoot`, which is the folder. So if the user only has access to the subfolder, the upload fails. Maybe we open a bug? 

- I don't think downgrading is likely an option at this point
- Creating new S3 buckets for each website is too much overhead
- Perhaps we look into another handler that doesn't have this problem (Digital Ocean?)

dleadbetter:
Ah, after actually pull thing, it appears that this PR does fix both problems with uploading? (I was a little confused by the description in the PR).

ajolipa:
> Ah, after actually pull thing, it appears that this PR does fix both problems with uploading? (I was a little confused by the description in the PR).

Oh yeah my bad, I didn't edit the description after pushing a fix for the second issue. I think the PR does now fix the issue, it just feels a bit brute force, but maybe that's okay for now? (And yeah, maybe we should open a bug...)

dleadbetter (commented):
I think this solution is fine for now, as it allows us to continue uploading to S3. Maybe let's open an issue in `next-tinacms-s3` just to track the progress and see if we can remove this work-around in the future. Could you:
- Update the PR description to indicate that it does solve the issue by working around a bug in `next-tinacms-s3`
- Add a comment to this endpoint explaining it is a workaround for an existing bug

PR #166, opened 2025-04-03
Use cached hits for map tooltip

camdendotlol:
# Summary

This PR updates the `Tooltip` component to look up the first `hits` item in the `useCachedHits` value.

It won't break anything to merge this on its own, but it's intended for use after https://github.com/performant-software/react-components/pull/483 is merged and that version of RC is pulled into CDP. That PR will remove the TS metadata fields that `Tooltip` relies on to get its label, necessitating the workaround in this PR.


dleadbetter (approved):

dleadbetter:
Could you connect this PR to an issue?

Issue #153, opened 2025-03-31
Static mode does not include place layers


dleadbetter:
When deploying the static mode, place layers do not display. Manifests are also not displaying.

## Server Mode
<img width="966" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNFY2Q1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--7776d7f73b0bb04f6566ffd1b701fdebd17c5659/Screenshot%202025-03-31%20at%207.43.16%E2%80%AFAM.png" alt="Screenshot 2025-03-31 at 7.43.16 AM.png" />

## Static Mode
<img width="932" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNFo2Q1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--4013bff8d8fabeae52732f2e4b0422d483e574bf/Screenshot%202025-03-31%20at%207.43.20%E2%80%AFAM.png" alt="Screenshot 2025-03-31 at 7.43.20 AM.png" />

PR #167, opened 2025-04-04
CDP #153 - Place layers

dleadbetter:
This pull request refactors some of the service logic in an attempt to be consistent with the way we're fetching data in both "server" and "static" modes. There were two issues that were present in "static" mode:

1. When fetching a place record, the nested `place_layers` were not present. This was due to the way we were fetching the data in data loader for Core Data. We were using the `GET /public/v1/:model` route instead of `GET /public/v1/:model/:uuid`. The later will typically include the "full" version of the record intended for detail pages, while the former is a more stripped down version intended for list pages.

2. When fetching data for any record, related manifests were not displaying. This was due to the way that related manifests are handled compared to other related records. In order to provide a valid IIIF manifest URL, the payload of the `/:model/:uuid/manfiests` endpoint is simply the IIIF manifest. Payloads for other related endpoints will using include a key as the base object. We were attempting to fetch both manifests and other related records the same way when the payload structures differed.

```
{
  [:model_name]: {
     ...attributes
  }
}
```

## Loaders
This pull request updates the loaders to the the`@performant-software/core-data/ssr` service classes to fetch the data. This eliminates the need to build URLs to request data from Core Data.

The loaders are now solely responsible for fetching data at build time. Services will not use any loader functions.

## Services
This pull request updates the services to be the source of truth for data. Services will make the determination if data should be fetched from the Astro content layer or Core Data. Server rendered pages will use services directly to fetch data. Client render pages will use services via an API.


dleadbetter:
@camdendotlol and @ajolipa - Tagging you both for a PR review as you've both recently done some work in the areas I refactored. Leaving this PR as a draft for now as it is dependent on a new release of `react-components` packages.

camdendotlol (approved):

Issue #152, opened 2025-03-28
Inconsistent margin on `Container`


dleadbetter:
It looks like the margins on the `Container` component are being applied inconsistently. With the screen width at 1535px, the margins looks fine.

<img width="1742" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNngxQ1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--5497e1cb20510d10b98c0a177dbe78b168dfc75c/Screenshot%202025-03-28%20at%207.03.34%E2%80%AFAM.png" alt="Screenshot 2025-03-28 at 7.03.34 AM.png" />

If I move to 1536px, the margins are huge.

<img width="1760" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNjUxQ1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--452f332c07687f060302e4847d3d3b4c4495aae2/Screenshot%202025-03-28%20at%207.03.46%E2%80%AFAM.png" alt="Screenshot 2025-03-28 at 7.03.46 AM.png" />

PR #173, opened 2025-04-07
CDP #152 - Post content

dleadbetter:
This pull request fixes a bug where the `margin: auto` CSS was causing an issue within a flexbox container for the post content. The solution was to add the `w-full` class to the container in order to set the width to "100%".

This pull request also fixes an issue with the map component not taking up the entire width of the container. To fix that, I removed the `mx-auto` class on the child containers and added `w-full`.

Finally, this pull request removes the horizontal padding on the Table visualization container to align the table with the header/content.

## Post page
Screenshot 2025-04-07 at 11 42 06 AM

## Post with table
Screenshot 2025-04-07 at 11 43 05 AM

### DIFF
--- a/src/apps/posts/PlaceInsert.tsx
+++ b/src/apps/posts/PlaceInsert.tsx
@@ -14,10 +14,10 @@ const PlaceInsert = (props: any) => {
 
   return (
     <div
-      className='flex flex-col gap-y-2 place-content-center mx-auto my-8 w-full'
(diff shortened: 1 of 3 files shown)


ajolipa (approved):
Assuming we're okay with having the embedded maps be full width (I think in the original designs they were more like 3/4 width but I could be wrong) then this looks good to me.

Issue #177, opened 2025-04-08
Timeline embed unavailable


dleadbetter:
Unless the `search.timeline.event_path` attribute is present, the Timeline embed shows an error. We should not require the `event_path` attribute for event based indexes.

<img width="720" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOXFZQ1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--e69481e15f13bf9dc041edd9e70ffa2bebfc8fd7/image.png" alt="image.png" />

PR #178, opened 2025-04-08
CDP #177 - Timeline embed

dleadbetter:
This pull request fixes an issue where the Timeline embed was not available in a post record unless the `event_path` configuration option was explicitly set, which should not be necessary for event-based indexes. The solution was to add a utility method to check whether to include the Timeline, which will be available to the two places that use it.

Additionally, this pull request fixes a layout issue with the posts index page.

Screenshot 2025-04-08 at 2 32 33 PM


blms (approved):
Looks good! I missed that you were checking this in both `TimelineInput.tsx`  and `visualizations.ts`, glad you caught it quickly.

PR #185, opened 2025-04-15
Cover images for detail pages

camdendotlol:
# Summary

This PR adds cover images to all models, and adds a related places map in the cover banner for non-place records.

When both a cover image and related places exist, the cover image takes up 30% of the width and the map takes up 70%. If only one is available, that component takes 100% of the width.

## Screenshot

Screenshot 2025-04-15 at 3 50 55 PM

### DIFF
--- a/src/pages/[lang]/places/[uuid]/index.astro
+++ b/src/pages/[lang]/places/[uuid]/index.astro
@@ -1,12 +1,11 @@
 ---
-import Layout from "@layouts/Layout.astro";
-import config from "@config";
-import RecordDetail from "@apps/pages/RecordDetail/index.astro";
+import Layout from '@layouts/Layout.astro';
+import config from '@config';
+import RecordDetail from '@apps/pages/RecordDetail/index.astro';
 import { getTranslations } from '@backend/i18n';
 import { ModelNames } from '@services/coreData/factory';
 import PlacesService from '@services/coreData/places';
-import PlaceMap from "@apps/pages/RecordDetail/PlaceMap";
-import { getDetailPagePaths } from '@utils/detailPages';
+import { getCoverImage, getDetailPagePaths } from '@utils/detailPages';
 
 const { t } = await getTranslations(Astro.currentLocale);
 
@@ -14,6 +13,9 @@ const { lang, uuid } = Astro.params;
 
 const { place } = await PlacesService.getFull(uuid);
 const excludes = config.search?.result_filtering?.places?.exclude || [];
+
+const coverUrl = getCoverImage(place);
+const geometry = place?.place_geometry?.geometry_json
 
 export const getStaticPaths = async () => await getDetailPagePaths(config, ModelNames.places);
 ---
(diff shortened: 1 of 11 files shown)


src/pages/[lang]/places/[uuid]/index.astro
Comment (jamiefolsom) [inline src/pages/[lang]/places/[uuid]/index.astro:1]: Does this present an image for place records only? Ideally, we would want to do pull and present a cover image for any record type.

src/pages/[lang]/places/[uuid]/index.astro
```diff
@@ -15,7 +15,35 @@ const { lang, uuid } = Astro.params;
 const { place } = await PlacesService.getFull(uuid);
 const excludes = config.search?.result_filtering?.places?.exclude || [];
 
-export const getStaticPaths = async () => await getDetailPagePaths(config, ModelNames.places);
+let coverUrl = null;
+
+const firstManifest = place.relatedRecords.manifests?.items;
+
+if (firstManifest && firstManifest.length > 0) {
```
Comment (dleadbetter) [inline src/pages/[lang]/places/[uuid]/index.astro:22]: The public Core Data API has a `/model/:uuid/media_contents` route that would probably be more straight forward to use than parsing the manifest. We'd also need to update the content loaders, services, and API.

camdendotlol:
@dleadbetter I've updated this to show cover images on all models, and a related places map on non-place pages.

The related places map can't be disabled with `excludeFields` because it's a combination of all related places instead of just one relationship. Not sure if that's an issue.

I think the panel cover image should be a separate PR since it will require `react-components` changes.

src/services/coreData/base.ts
```diff
@@ -187,6 +189,24 @@ class Base {
     return manifests;
   }
 
+  /**
+   * Returns the related media contents for the record with the passed ID.
+   *
+   * @param id
+   */
+    async getRelatedMediaContents(id: string) {
```
dleadbetter, line 197: Do we also need to update the loaders for this to work in "static" mode?

src/utils/detailPages.ts
```diff
@@ -19,4 +20,20 @@ export const getDetailPagePaths = (async (config: any, model: Models) => {
   }
 
   return routes;
-});
\ No newline at end of file
+});
+
+export const getCoverImage = (record: any) => {
+  if (record?.relatedRecords?.mediaContents && record.relatedRecords.mediaContents.length > 0) {
+    return record.relatedRecords.mediaContents[0].content_url
```
dleadbetter, line 27: The `content_url` property will load the image from Core Data's Active Storage backend. We'll probably want to load it using the IIIF Image API, for which a few different properties should be available:

`content_iiif_url` - Full image
`content_preview_url` - Max width of 500px
`content_thumbnail_url` - Max width of 250px

camdendotlol: I've replaced `content_url` with `content_iiif_url`.

src/services/coreData/base.ts
dleadbetter: Looks like #186 implements the Core Data loaders, can we pull `develop` into this branch once that PR is merged and confirm that this will work in "static" mode?

camdendotlol: Done!

dleadbetter (approved):

PR #186, opened 2025-04-17
Cover image for detail panels

camdendotlol:
# Summary

This PR uses the new `coverUrl` prop from https://github.com/performant-software/react-components/pull/490 to enable cover images in the search view. Like the detail pages in #185, it takes the first related media content record in the list, which comes pre-sorted by Core Data.

## Possible issues

In the recording below you can see that the images take a couple seconds to load, even on my fast internet. The flash as the image loads in is a bit jarring. It might be better to use the manifest thumbnail if it's not too low-res and if we're certain that IIIF Cloud is setting the thumbnail to the image with the lowest `order` value.

## Screen recording

https://github.com/user-attachments/assets/e3d6ef82-fd96-4059-b2c3-c12531eb58a7


src/apps/search/panels/BasePanel.tsx
```diff
@@ -165,6 +171,17 @@ const BasePanel = (props: Props) => {
   },
   [data, props.name]);
 
+  /**
+   * Memo-izes the cover URL.
+   */
+  const coverUrl = useMemo(() => {
+    if (!mediaContentsLoading && mediaContents.length > 0) {
+      return mediaContents[0].content_url
```
dleadbetter, line 179: See [comment](https://github.com/performant-software/core-data-places/pull/185/files#r2050702733) about `content_url` in #185. Addiontally, to address any loading lag, we could render a gray placeholder.

camdendotlol: Would it be alright to show the placeholder for records that don't have any images?

If not, the placeholder would flash and then disappear for those records because we can't know whether a record has images until we load from the `media_contents` endpoint.

dleadbetter: Hm, that's a good question. Could we render the placeholder after we've loaded the records from the `media_contents` endpoint? I suspect the request from the `<img` tag is what is taking a long time.

camdendotlol: Hm, so we would watch the `complete` attribute of the `` element and display the placeholder until `complete = true`?

camdendotlol: The main issue above I think is the layout shift where the image element is only inserted once we know the record has associated media. That means the rest of the panel's content jumps up and down as we move between records that have media and those that don't.

camdendotlol:
@dleadbetter

I've updated the behavior and I think it feels a bit nicer to use now. When the user clicks on a new record:

* if the previous record had an image, display the gray placeholder PNG until we've loaded the `media_contents` for the new record (this prevents the double layout shift when clicking between records with media contents)
* if the previous record did *not* have an image, do not set a cover image URL until the new record finishes fetching its media contents (like above, this means there's no sudden layout shift when we switch records)
* when we've finished fetching the list of media contents:
  * if no media contents exist, set `coverUrl` to `null` so the cover disappears
  * if media records exist, use the `content_preview_url` of the first one

src/apps/search/panels/BasePanel.tsx
```diff
@@ -165,6 +173,25 @@ const BasePanel = (props: Props) => {
   },
   [data, props.name]);
 
+  /**
+   * Updates the cover URL when the record or loading states change.
+   */
+  useEffect(() => {
```
dleadbetter, line 179: Nice, I like this approach.

src/services/coreData/base.ts
```diff
@@ -169,6 +169,24 @@ class Base {
     return { items };
   }
 
+  /**
+   * Returns the related media contents for the record with the passed ID.
+   *
+   * @param id
+   */
+    async getRelatedMedia(id: string) {
```
dleadbetter, line 177: We'll also need to update the content loaders to fetch related media contents for this to work in static mode.

src/services/coreData/factory.ts
```diff
@@ -12,6 +12,7 @@ export const ModelNames = {
   events: 'events',
   instances: 'instances',
   items: 'items',
+  mediaContents: 'media_contents',
```
dleadbetter, line 15: Do we need this if we're only ever fetching related media records?

dleadbetter (approved):

Issue #208, opened 2025-05-13
Bug: Tina media insert not working


ajolipa:
On the most recent version, the API call made by the `TinaMediaPicker` component returns a `404` response, causing the component to be stuck in a loading state forever.

PR #209, opened 2025-05-13
Adding `media_contents` to services factory

ajolipa:
### In this PR
Addresses Issue #208 by adding the missing `media_contents` case to `services/coreData/factory.ts`.


dleadbetter (approved):
Good catch!

Issue #213, opened 2025-05-20
Documentation


dleadbetter:
- [ ] Refactor build process
  - Currently there are several build scripts which are all run separately
  - Let's refactor everything into a single build script
  - Individual functions can still live in their own files
- [ ] Add additional documentation for `config.json`
- [ ] Add documentation for upgrade notes

PR #215, opened 2025-05-23
CDP #213 - Documentation

dleadbetter:
This pull request makes the following changes to documentation and the build process:

- Re-organizes the build scripts
- Adds Vitest
  - Adds test for `config.json` to build process
- Refactors documentation and `README.md`
  - Re-organizes the documentation files
  - Adds additional documentation for `config.json`
  - Adds documentation for upgrade notes (previously in Slack)


dleadbetter:
@blms, @camdendotlol, and @ajolipa - Adding you all as reviewers, as I believe you've all done Core Data Places deployments. If there is anything missing from the documentation that you think would be helpful for either setting up a new Core Data Places site or modifying `config.json` for an existing site, please let me know and I will add it to the documentation. Thanks!

src/types.ts
```diff
@@ -5,47 +5,59 @@ export interface SearchConfig {
...
+    collections?: Array<String>,
+    localize_pages?: boolean
   };
 
-  content?: Array<'posts' | 'paths'>;
-
   core_data: {
     url: string,
-    project_ids: number[]
+    project_ids: string[]
```
ajolipa, line 60: Does this match the existing GBoF projects? I think on most/all of them the project IDs are in the config as numbers -- will that cause any issues with the config tests?

ajolipa (approved):
One tiny question re: types (see comment) but looks good!

src/types.ts
dleadbetter: Yes, it likely will cause an issue with existing `config.json` files if the project IDs are numbers. The problem is that TinaCMS doesn't allow `number[]` as a type, so we have to use `string[]` in the TinaCMS schema. If the existing value is an array of integers and the settings are modified via Tina, they will get converted to an array of strings anyway.

When building the request URLs, it doesn't seem to matter if these values are numbers or strings (I think they likely get coerced into strings at some point anyway), so it's probably worth us being consistent in entering them as `string[]`.

ajolipa: Yeah that all makes sense! Just might require updating a handful of existing configs.

camdendotlol (approved):

PR #228, opened 2025-07-30
Fixing null geometry detail panel crash

ajolipa:
### In this PR
Adds some handling for the case where places don't have defined geometry to the `BasePanel` and `Place` panel components, to avoid the page crashing a place without geometry is clicked on in the results list.

### Notes
This is a bit of a bandaid obviously; it might be better to handle this case more smoothly in the `CoreDataUtils.toFeatureCollection` function, instead of always having to make sure not to feed that method a place without geometry data?


dleadbetter:
> This is a bit of a bandaid obviously; it might be better to handle this case more smoothly in the CoreDataUtils.toFeatureCollection function, instead of always having to make sure not to feed that method a place without geometry data?

Yes, let's handle this in `CoreDataUtils` so that we only have to do the check in one place.

PR #235, opened 2025-08-26
Adding author and date fields to paths/posts

ajolipa:
### In this PR
Adds optional `author` and `date` fields to the `paths` and `posts` models, and adds them to the `Card` component and to the individual path/post pages. 
image
image

### DIFF
--- a/src/apps/posts/PostContent.tsx
+++ b/src/apps/posts/PostContent.tsx
@@ -16,7 +16,7 @@ interface PostContentProps {
   content: TinaMarkdownContent;
   title: string;
   author?: string;
-  date?: string;
+  date?: string | number | Date;
 }
 
 const PostContent = (props: PostContentProps) => {
@@ -39,7 +39,7 @@ const PostContent = (props: PostContentProps) => {
             >
               { props.title }
             </h1>
-            { (props.author || props.date) && <Byline author={props.author} date={props.date} t={t} /> }
+            { (props.author || props.date) && <Byline author={props.author} date={props.date} /> }
             <article
               className='prose prose-lg max-w-none w-full'
             >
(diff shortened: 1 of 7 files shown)


src/apps/posts/PostContent.tsx
```diff
@@ -36,6 +39,7 @@ const PostContent = (props: PostContentProps) => {
             >
               { props.title }
             </h1>
+            { (props.author || props.date) && <Byline author={props.author} date={props.date} t={t} /> }
```
dleadbetter, line 42: Since this component is nested within the `TranslationContext`, you should be able to use the context to retrieve the `t` function instead of passing it as a prop:

```
```

src/pages/[lang]/paths/index.astro
```diff
@@ -38,6 +38,8 @@ export const getStaticPaths = () => _.map(config.i18n.locales, (lang) => ({ para
           slug={getRelativeLocaleUrl(Astro.currentLocale, `/paths/${path?._sys.filename}` || '#')}
           t={t}
           title={path?.title || t('untitled')}
+          author={path?.author}
+          date={path?.date ? new Date(path.date).toLocaleDateString() : null}
```
Comment (dleadbetter) [inline src/pages/[lang]/paths/index.astro:42]: Is the date always going to be a string that is parsed into a date and formatted? Would it make sense to handle that business logic in the `Card` and `Byline` components?

src/apps/posts/PostContent.tsx
ajolipa: Ah great point.

src/pages/[lang]/paths/index.astro
Comment (ajolipa) [reply src/pages/[lang]/paths/index.astro:42]: Yeah that makes sense, especially since we might want it formatted differently in different contexts.

dleadbetter (approved):

PR #249, opened 2025-09-18
Tina Visual Editing

ajolipa:
## In this PR
This is a work in progress to implement visual editing for user-generated content in TinaCMS. This branch currently implements it for both paths and posts, and each of these has one issue I haven't been able to figure out. I'm including both here since I'm not sure which will turn out to be the most tractable.

- Posts: This is the more serious issue but also the one I suspect is an easier fix. When editing the body of a post above an inserted map component, the page crashes with the following console errors:
```
Uncaught TypeError: can't access property "removeLayer", this.style is undefined
    removeLayer maplibre-gl.js:46
    g21 Map.js:74
    o26 WarpedImageLayerPeripleo.js:46
    React 14
    workLoop scheduler.development.js:266
    flushWork scheduler.development.js:239
    performWorkUntilDeadline scheduler.development.js:533
    js scheduler.development.js:571
    js scheduler.development.js:633
    __require chunk-DWA4UIM3.js:12
    js index.js:6
    __require chunk-DWA4UIM3.js:12
    React 2
    __require chunk-DWA4UIM3.js:12
    js React
    __require chunk-DWA4UIM3.js:12
    js React
    __require chunk-DWA4UIM3.js:12
    <anonymous> client.js:2
```
and a bunch of similar things. I believe the issue has to do with how the `PlaceInsert` component is being unmounted when the data passed to the body of the post changes so that the map is in a new position in the list of children, but I haven't been able to figure out how to make it handle this state change more elegantly.

- Paths: The issue here I find so far completely baffling; when the editing sidebar is present, clicking on any interactive part of the path viewer seems to just make Tina think you're trying to add a new place to the path:

https://github.com/user-attachments/assets/df451fb4-a8aa-4032-8f0a-9d627eb7ee2d

This can be worked around by collapsing the side panel and navigating to the desired leg of the path and then reopening the panel, but that is not an ideal user experience.


rsimon:
Hi @ajolipa, I saw this just now. Looks like it could be some kind of race condition inside that Peripleo component's effect hooks. I'll take a look!

rsimon:
Hi @ajolipa, I found a likely culprit. I don't have a full setup of `core-data-places` locally, so couldn't test this end-to-end. But I did get a very similar error with a local test setup, where I unmounted a map with a raster layer.

I'm going to publish a new Peripleo release, which fixed the problem in my test case. So, one thing you could try is to update to new Peripleo. (Will be v0.8.8).

**However:** I think you are not running into this problem inside Peripleo, but actually earlier. From the logs above, it looks like it happens here:

https://github.com/performant-software/react-components/blob/09dd32bdd9b42e068799e8975ba9e5e26239c9ca/packages/geospatial/src/utils/Map.js#L74

If that's the actual error location, you'd have to apply a similar fix here as I've applied in Peripleo:

https://github.com/peripleo/peripleo/blob/f2c7fdfbb3ea1352d3c4dc0175a67626f073d378/packages/peripleo-maplibre/src/map/utils.ts#L145-L157

It's kind of odd, and definitely a race condition. Here's my guess of what's going:
- When you unmount the React component, the MapLibre map is in the middle of the unmounting process.
- At this point, the  `map` object still exists. But, internally, some parts are already destroyed. That's specifically the case for the `style` (according to your error log; and I also confirmed the same thing in my test setup).
- Therefore - although not ideal - I'm now testing whether `map.style` exists, rather than just `map`. In my example, I added an additional catch block. But that shouldn't get used anymore (at least not in this case), since the `map.style` check seemed to do the trick.

PR #258, opened 2025-09-25
Improving data loader performance

ajolipa:
### In this PR
This PR attempts to make progress on improving error handling and adding more flexible build options for different needs, in the following ways:
- Wrapping the API calls for each record in a `try...catch` block so that the whole build process (hopefully) won't abort if there's an error fetching the data for one record;
- Adding a 10ms delay between fetching the data for each record, to (hopefully) avoid making the FairData server think it's under a DDOS attack or something;
- Adding an environment variable `CONTENT_MODE` that can be set to `add` if the desired behavior is to avoid making API calls for any records with existing entries in the cached content layer; this can be used in the case of an aborted build or if a subset of records were not fetched successfully in one build run;
- Alternatively if `CONTENT_MODE` is set to `refresh`, the data store will be fully cleared before the new data is fetched; this allows for a full refresh in the case that records were deleted between builds (otherwise deleted records are not removed from the data store cache);
- If `CONTENT_MODE=update` (or is unset), then at build time all data will be fetched (necessary to check for changes), but the data store entry will only be updated if the data for that record has changed (this is the current behavior).
- I also added a `meta.lastModified` field to the data store, which I admit we're not currently using for anything but I figure it can't hurt? Could give useful information for diagnosing build errors or making sure things were really updated.

### DIFF
--- a/src/loaders/coreData/helpers.ts
+++ b/src/loaders/coreData/helpers.ts
@@ -52,6 +52,7 @@ export const createLoader = (name, Service, [indexName, showName]) => ({
     let count = 1;
     let pages = 1;
     let total = 1;
+    let failed = 0;
 
     const records = [];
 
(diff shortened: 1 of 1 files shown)


blms (approved):
Looks great! I like the different modes idea, and the `add` mode is a clear and effective efficiency improvement.

PR #276, opened 2025-10-20
Adding stacked comparison timeline visualization

ajolipa:
### In this PR
Adds a data visualization for `event` models that displays the uploaded set of search results as stacked bars representing the date range of each event.
image

### Notes and questions
I feel like I could spend infinite time trying to refine the styling of this and add flexibility and more customization, but I figured I should get a PR up at this stage and get some feedback. Some notes:
- For simplicity I only made this component work with search results from actual `event` models for now, but I suppose in theory it could be expanded to any search that has a related `event` model? Would require configuration in the `config` file a la the `timeline` config, so I held off on that for now, but could add if desired.
- We could include another categorical dimension of information in the chart by coloring the bars, e.g. if the user wanted to include events related to two different regions in the same chart and distinguish them by color. If we are doing configuration, this field could also be configured there.


blms:
Looking good so far! A few thoughts:

- Could we simplify the tooltip to just be `Event Name (Date–Range)` instead of having the date range field on a separate line and with a label in all caps?
- We should decide how to handle single dates (i.e. the start and end is the same, or there is no end date). Right now they display as labels but without anything on the graph. Maybe just make them have minimum one pixel width as a very basic patch? (If that's possible)
- It also looks like it's interpreting "no end date" as 1969 (epoch I assume).

Screenshot 2025-10-21 at 4 09 34 PM

ajolipa:
> Could we simplify the tooltip to just be `Event Name (Date–Range)` instead of having the date range field on a separate line and with a label in all caps?

Definitely possible; to keep things simple I didn't implement a custom tooltip yet but I agree it would be preferable, and should be quick to make.

> We should decide how to handle single dates (i.e. the start and end is the same, or there is no end date). Right now they display as labels but without anything on the graph. Maybe just make them have minimum one pixel width as a very basic patch? (If that's possible)

Hmm good catch. I tried to cheese this a bit by filtering out events that lack either a start or end date from the data when initially transforming it in Tina, but of course that can still be edited manually and also doesn't deal with the case of start and end dates equal, so yeah, more checks need to happen at the stage of actually rendering the graph. 

> It also looks like it's interpreting "no end date" as 1969 (epoch I assume).

What do you think would make the most sense to handle this case? Just assume end date = start date in this case? Exclude the event from the visualization because of missing info?

blms:
> What do you think would make the most sense to handle this case? Just assume end date = start date in this case? Exclude the event from the visualization because of missing info?

I would say just assume end date = start date, that's how we handle it in other places. And while those generally aren't good matches for this type of timeline, we probably can't prevent users from using them here—and better to show something than nothing!

blms:
The custom tooltip and min width look great, and thanks for upgrading recharts!

I'm seeing that you also allowed using the timeline config when Event isn't the main model; however, I'm getting an error on the scavenger hunt data when trying to use an export from Places (which does have [the `event_path` configured](https://raw.githubusercontent.com/performant-software/pss-scavenger-hunt-content/refs/heads/main/content/settings/config.json)), `Cannot read properties of undefined (reading 'start_date')`. Could it be that it's checking every record for dates? 

(Also the Clear button is giving me `Cannot destructure property 'name' of 'JSON.parse(...)' as it is null.`)

I also wonder if we somehow need to communicate to users that they can only use this for events, or when the timeline is configured in the config… maybe not in scope for this work but I can foresee it being problematic in the future.

ajolipa:
> I'm seeing that you also allowed using the timeline config when Event isn't the main model; however, I'm getting an error on the scavenger hunt data when trying to use an export from Places (which does have [the `event_path` configured](https://raw.githubusercontent.com/performant-software/pss-scavenger-hunt-content/refs/heads/main/content/settings/config.json)), `Cannot read properties of undefined (reading 'start_date')`. Could it be that it's checking every record for dates?

Hmm, possibly.... my guess is this is an easy fix and like you said it's just looking somewhere it shouldn't be looking. I shall investigate.

> (Also the Clear button is giving me `Cannot destructure property 'name' of 'JSON.parse(...)' as it is null.`)

Weird!! Will check it out.

> I also wonder if we somehow need to communicate to users that they can only use this for events, or when the timeline is configured in the config… maybe not in scope for this work but I can foresee it being problematic in the future.

Yeah, for now it just gives an error message if you try to upload JSON data that doesn't have timeline configured and isn't an event model, but it could also be added to the description pretty easily.

PR #282, opened 2025-10-24
List Search

camdendotlol:
# Summary

## List search

This PR adds a new type of mapless search page to Core Data Places. Using a new `type` field in the search config, users can specify that they want a list or grid layout instead. These layouts use a separate instance of InstantSearch that is decoupled from the Peripleo map logic, but it does share the same `Facets` component. (If `type` is empty, e.g. in the case of projects configured with older CDP releases, map search will be assumed.)

There are no (intentional) breaking changes to the config.

## Custom components

This PR also adds a new custom components feature.

Custom components must be placed in `/components/[name].tsx` in the project content repo where `[name]` is the name of the default component. The component must support the same set or a subset of the default component's props (i.e. every prop expected by the custom component must be accepted by the default component). This PR adds a test to the Vitest config to enforce this rule.

The list search makes use of customizable `ListHit` and `GridHit` components.


scripts/build.mjs
```diff
@@ -19,4 +20,7 @@ import { buildSearch } from './build.search.mjs';
 
   console.log('Building search.json from configuration...');
   await buildSearch(config);
+
+  console.log('Fetching custom components...')
+  copyComponents()
```
dleadbetter: Should this include `await`? Missing semi-colons.

src/apps/search/list/ListSearch.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+interface Props {
+  name: string;
+  searchConfig: SearchConfig;
+}
+
+const ListSearch = (props: Props) => {
+  const { typesense } = props.searchConfig;
+
+  const searchClient = useMemo(() => new TypesenseInstantsearchAdapter({
```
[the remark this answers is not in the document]
dleadbetter: May also want to use [creatRouting](https://github.com/performant-software/react-components/blob/e69e8424ed76bb1609eabf56f909c6191b15a82a/packages/core-data/src/utils/Typesense.js#L45) if we want to sync the search state with the URL.

src/apps/search/list/ListSearchContext.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+const ListSearchContext = createContext<SearchContextType>(null);
+
+interface Props {
+  children: ReactNode;
+  name: string;
+}
+
+export const ListSearchContextProvider = ({ children, name }: Props) => {
+  const searchConfig = useMemo(() => _.findWhere(config.search, { name }), [config, name]);
```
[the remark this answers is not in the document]
dleadbetter: Or perhaps we use a separate context for config vs. the other stuff used for maps?

src/components/custom/default/ListHit.tsx
```diff
@@ -0,0 +1,109 @@
+import PropTypes from "prop-types";
+import {useMemo} from "react";
+import {Pill} from "@performant-software/core-data/ssr";
+
+interface Props {
```
dleadbetter: Should we factor this out somewhere common if it's used for multiple components?

src/apps/search/list/Hits.tsx
```diff
@@ -0,0 +1,143 @@
+import {useHits} from 'react-instantsearch';
+import ListHit from '@components/custom/project/ListHit'
+import {useTranslations} from "@i18n/useTranslations";
```
dleadbetter: The `useTranslation` hook will make a `fetch` request to the API to retrieve the translations from Tina. This should only be done once at the most top level component. We can then use `TranslationContext` to supply the translation values to the component further down the component hierarchy.

dleadbetter (commented):
Looks good. Could we make one more pass through to make sure the formatting is consistent? Mainly single quotes instead of double quotes, use of semi-colons, and spaces before/after curly braces.

dleadbetter (approved):

PR #288, opened 2025-11-06
Disable facet and search input when map results are still loading

camdendotlol:
# Summary

This PR fixes the root cause of the state management issue described in the comments of https://github.com/performant-software/pan-african-data-content/issues/4 by disabling the facet button and the search query input while the initial results are still loading.

Since the pages are being fetched automatically in the background on a 25ms timeout, we can't rely directly on the `status` field of `useInstantSearch`, because it will briefly say that it's finished loading between page requests. So this PR implements a timeout where we only enable the search after `status` hasn't indicated that anything is loading for a period of 250ms.

This logic only applies if the search query is empty (i.e. on first load). We don't want to freeze the input box while the user is typing. That should be okay, because searches very rarely result in more than one page of results.

### DIFF
--- a/src/apps/search/Header.tsx
+++ b/src/apps/search/Header.tsx
@@ -13,7 +13,7 @@ import clsx from 'clsx';
 import { useContext, useMemo } from 'react';
 import { useCurrentRefinements } from 'react-instantsearch';
 import _ from 'underscore';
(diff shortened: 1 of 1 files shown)


camdendotlol:
Done! I vaguely remembered that hook but couldn't find it when I looked for it.

PR #291, opened 2025-11-06
Updating fill color of selected region

ajolipa:
### In this PR
Updates the fill style passed to the `LocationMarkers` component in `BasePanel` to make it distinct from the default. This avoids visual ambiguity/confusion between overlapping regions and a selected/highlighted region.

image

### Notes
I tried not to go down a rabbit hole about finding the perfect color haha; I tried some blue-ish colors and didn't love how they looked, and to my eye the gold looks harmonious and reads easily as "highlight". But of course I am open to suggestions from others about the exact color value here!

### DIFF
--- a/src/apps/search/panels/BasePanel.tsx
+++ b/src/apps/search/panels/BasePanel.tsx
@@ -1,26 +1,30 @@
-import ManifestThumbnail, {
-  type Collection,
-} from "@apps/search/ManifestThumbnail";
-import SearchContext, { useSearchConfig } from "@apps/search/SearchContext";
-import Base from "@backend/api/coreData/base";
-import UserDefinedFieldView from "@components/UserDefinedFieldView";
-import TranslationContext from "@contexts/TranslationContext";
+import ManifestThumbnail, { type Collection } from '@apps/search/ManifestThumbnail';
+import SearchContext, { useSearchConfig } from '@apps/search/SearchContext';
+import Base from '@backend/api/coreData/base';
+import UserDefinedFieldView from '@components/UserDefinedFieldView';
+import TranslationContext from '@contexts/TranslationContext';
 import {
   CoreData as CoreDataUtils,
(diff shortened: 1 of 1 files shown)


dleadbetter (commented):
Woah, there's some serious auto-formatting going on here. Any chance we can not do the auto-formatting? It'll make the diff a lot easier to read.

src/apps/search/panels/BasePanel.tsx
```diff
@@ -442,23 +567,26 @@ const { data: { people = [] } = {}, loading: peopleLoading } = useLoader(onLoadP
...
         <LocationMarkers
           animate
           boundingBoxOptions={boundingBoxOptions}
           data={geometry}
-          fitBoundingBox={_.get(config.map, 'zoom_to_place', true)}
-          layerId='current'
+          fillStyle={{
+            type: "fill",
+            paint: {
+              "fill-color": "#ffd546",
```
jamiefolsom: To me it seems as though this should be configurable.

src/apps/search/panels/BasePanel.tsx
```diff
@@ -442,23 +567,26 @@ const { data: { people = [] } = {}, loading: peopleLoading } = useLoader(onLoadP
...
+              "fill-opacity": 0.4,
```
jamiefolsom: Probably this too.

src/apps/search/panels/BasePanel.tsx
ajolipa: Consensus on Slack seemed to be not to overcomplicate by adding to the config, although it's certainly possible if that's what we want. Although the color of the base fill and points isn't configurable, so it would be a bit weird if only this highlight color were.

dleadbetter (approved):
This looks good to me. I agree that having the highlight color configurable, but not the other map layers probably isn't ideal.

PR #293, opened 2025-11-07
CDP #287 - Place picker

dleadbetter:
This pull request updates the `/pages/api/[model]/index.json` API endpoint to parse the URL search parameters on the request and forward that to the data fetching service. For server rendered sites, this will add the parameters to the request sent to the Fair Data API. For static rendered sites, this will have no affect, as all of the data is stored in static JSON files and runtime filtering is not possible.

This pull request also refactors the `TinaPlacePicker`, `TinaModelPicker`, and `TinaMediaPicker` components to fetch the first 25 records from the API when the component is mounted. The user will then be required to type into the search input in order to retrieve the record they wish to select.

Screenshot 2025-11-06 at 8 47 24 PM
Screenshot 2025-11-06 at 8 47 40 PM

Lastly, this pull request also adds the optional `CONFIG_FILE` environment variable to be used in development environments to specify which config file to copy into `./public/config.json`.


dleadbetter:
> If the static site is hosted on Netlify then I believe the Tina backend stuff will still work, e.g. https://static.panafricandata.org/admin/ . Clearly for a static version of the site hosted on GHPages or similar it won't, but I think on Netlify the way it's currently set up the Tina functionality isn't tied to the build mode in the Astro config.

It will certainly still work on Netlify, but if the pages are all generated at build time, none of the content added after will be visible. We generally tell projects that Tina is not available in static mode.

PR #318, opened 2025-11-17
Server islands

camdendotlol:
# Summary

This PR refactors the record detail and posts to use [Astro server components](https://docs.astro.build/en/guides/server-islands/) and the `server:defer` directive. The TL;DR is that each component should be responsible for fetching its own data wherever it makes sense to do so, because `server:defer`ed Astro components will load in after the initial page load instead of blocking the initial page load.

Some general notes:

* I removed the `getFull` method from the models' base class because it is no longer being used
* `GeoHeader` is an odd component because the related places need to be used in two different components, so I encapsulated them into this one component in order to avoid duplicate fetch requests. There is more stuff inside `GeoHeader`, such as the top-level record metadata, that needs to remain in that spot to avoid significant changes to the page layout.
* The initial page load is still blocked by the top-level `getOne` call because we need the name of the record to pass to the `Layout` component. That's not ideal, but there isn't really a way around it unless we want to use a generic page/tab title and then dynamically update it when the record loads in. I'm open to doing that but it would feel a bit janky.
* I inadvertently fixed a bug where we were trying to set the title of the person detail page with `person.name`, which led to the URL showing in the tab instead of the person's name (which is constructed with first name, last name, etc.)

### DIFF
--- a/src/apps/pages/RecordDetail/Relations.astro
+++ b/src/apps/pages/RecordDetail/Relations.astro
@@ -11,14 +11,14 @@ interface Props {
   uuid: string;
 }
 
-const { data, icon, label, resolveDetailUrl, resolveName } = Astro.props;
+const { data, icon, label, resolveDetailUrl, resolveName, showLinks } = Astro.props;
 ---
 {(Object.keys(data)).length > 0 && (
   <div class='py-6'>
     <h2 class='capitalize text-lg font-semibold mb-2'>{label} ({data.length})</h2>
     <ul class='grid grid-cols-3 items-start justify-start'>
       {data.map((record) =>
-        resolveDetailUrl ? (
+        showLinks && resolveDetailUrl ? (
           <li>
             <a
               class='flex gap-2 my-2 text-wrap items-center font-semibold'
--- a/src/services/coreData/base.ts
+++ b/src/services/coreData/base.ts
@@ -2,7 +2,6 @@ import config from '@config';
 import { hasContentCollection } from '@root/src/content.config';
 import { getCollection, getEntry } from 'astro:content';
 import _ from 'underscore';
-import { CoreData as CoreDataUtils } from '@performant-software/core-data/ssr';
 
 const REQUEST_PARAMS = {
   per_page: 0


src/apps/pages/RecordDetail/GeoHeader.astro
```diff
@@ -0,0 +1,58 @@
...
+  model: string
+}
+
+const { excludes, lang, record, model, name } = Astro.props;
+
+const service = ServiceFactory.getService(model);
+const { places } = await service.getRelatedPlaces(record.uuid, true);
+const { mediaContents } = await service.getRelatedMediaContents(record.uuid);
+
+const geometry = model === 'places'
```
dleadbetter: I'm not sure we want to be checking `model === <value>` in these child components, I think it could end up getting very messy. Could we instead include that business logic as a prop that is passed to this component?

src/apps/pages/RecordDetail/index.astro
```diff
@@ -1,169 +1,90 @@
...
+import Container from '@components/Container.astro';
+import RelatedEvents from '@apps/pages/RecordDetail/RelatedEvents.astro';
+import RelatedItems from '@apps/pages/RecordDetail/RelatedItems.astro';
+import RelatedInstances from '@apps/pages/RecordDetail/RelatedInstances.astro';
+import RelatedOrganizations from '@apps/pages/RecordDetail/RelatedOrganizations.astro';
+import RelatedPeople from '@apps/pages/RecordDetail/RelatedPeople.astro';
+import RelatedWorks from '@apps/pages/RecordDetail/RelatedWorks.astro';
+import RelatedTaxonomies from '@apps/pages/RecordDetail/RelatedTaxonomies.astro';
+import RelatedMedia from '@apps/pages/RecordDetail/RelatedMedia.astro';
+import GeoHeader from "@apps/pages/RecordDetail/GeoHeader.astro";
```
dleadbetter: Single quotes.

src/apps/pages/RecordDetail/GeoHeader.astro
camdendotlol: We can't pass functions as props to `server:defer` components because the props are serialized and sent to the server, so any `onLoad` or `getGeometry` type of thing isn't possible (this is also why we're looking up the service via the factory multiple times).

dleadbetter: For the place model, could we provide the geometry as a prop? Then do some sort of logic like: 

`geometry || getRelatedGeometry(places)`?

This way the places page is still responsible for the place-specific business logic.

dleadbetter: I think fetching the related places within the `GeoHeader` component is fine. I'm suggesting we move the `model === 'place'` specific logic _outside_ of this component  onto the place page, and pass the required data as a prop.

Issue #292, opened 2025-11-06
Visual confusion with overlapping vs highlighted regions


ajolipa:
In the case where the map contains many polygons, some of which overlap or are contained in each other, there's no visual indicator that differentiates between two regions on top of each other and a highlighted/selected region from the detail panel data -- basically in both cases we're just piling two 20% opacity fills of the same color on top of each other.

PR #320, opened 2025-11-18
MapFeaturesContext: Architecture Concept

rsimon:
## In this PR

This is a **draft PR** that proposes centralizing all map-rendered geo-features into a shared, context-driven state. The goal is to maintain the current behavior (auto-populating the map from TypeSense hits) while enabling other components to update individual places at runtime (e.g. when the `BasePanel` fetches higher-resolution feature data).

#### Key Changes

- Introduced a new `MapFeaturesContext` that internally uses `useCachedHits` to derive a shared `places` state.
- Added a `useCachedPlaces` hook that exposes:
  - the current list of places
  - an `updatePlace` method for replacing an existing place with an updated (higher-res) version
- Wrapped the main interface in `MapFeaturesContextProvider` within `MapLayout`.
- Updated `MapView` to use  `useCachedPlaces` instead of `useCachedHits`.
- Removed the additional `LocationMarkers` in `BasePanel` that previously created duplicate overlays. Instead, `BasePanel` now updates the shared places state via `updatePlace`.

**IMPORTANT**: this is all conceptual and untested, since I'm lacking a proper setup (TypeSense access credentials, Tina backend etc.) Consider this a "pseudo-code-ish" proposal intended to illustrate the idea, and for further discussion.

### DIFF
--- a/src/apps/search/map/panels/BasePanel.tsx
+++ b/src/apps/search/map/panels/BasePanel.tsx
@@ -26,7 +26,6 @@ import {
 import _ from 'underscore';
 import PanelHistoryContext from '@apps/search/map/PanelHistoryContext';
 import { useSearchConfig } from '@apps/search/SearchConfigContext';
-import { preProcessFile } from 'typescript';
 import { useCachedPlaces } from '../MapFeaturesContext';
 
 interface Props {
--- a/src/apps/search/map/panels/BasePanel.tsx
+++ b/src/apps/search/map/panels/BasePanel.tsx
@@ -26,6 +26,8 @@ import {
 import _ from 'underscore';
 import PanelHistoryContext from '@apps/search/map/PanelHistoryContext';
 import { useSearchConfig } from '@apps/search/SearchConfigContext';
+import { preProcessFile } from 'typescript';
+import { useCachedPlaces } from '../MapFeaturesContext';
 
 interface Props {
   className?: string;
@@ -49,6 +51,7 @@ const BasePanel = (props: Props) => {
 
   const navigate = useNavigate();
   const config = useSearchConfig();
+  const { updatePlace } = useCachedPlaces();
   const { t } = useContext(TranslationContext);
   const { setSelected } = useSelection();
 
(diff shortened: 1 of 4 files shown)


src/apps/search/map/panels/BasePanel.tsx
```diff
@@ -220,6 +223,19 @@ const { data: { people = [] } = {}, loading: peopleLoading } = useLoader(onLoadP
         places.filter((place) => place.place_geometry)
       );
   }, [item, places, props.resolveGeometry]);
+  */
+
+  useEffect(() => {
+    if (props.resolveGeometry && item) {
+      const highRes = props.resolveGeometry(item);
+      updatePlace(highRes);
+    } else if (!_.isEmpty(places.filter((place) => place.place_geometry))) {
```
rsimon: TBH I didn't figure out what this branch does. Under what conditions will new places come in? I assume this code needs changing... instead of creating one FeatureCollection, we probably need to loop through each place and call `updatePlace` for each one separately?

jamiefolsom:
@dleadbetter would you have a look at this proposal, and comment, and if you like the approach, provide @rsimon with access to actually implement it?

EDIT:

By way of context, this is the first step in trying to address the issues seen on the Pan African Data site, where there are several overlapping layers of polygons of differing resolutions.

Screenshot 2025-11-18 at 16 10 02

Screenshot 2025-11-18 at 16 10 37

dleadbetter (commented):
I think this implementation is likely a step in the right direction, but I don't think it is going to completely solve the issue with overlapping polygons (and maybe that's okay/expected?): Although we're replacing the geometry of the selected record with a higher resolution polygon, adjacent polygons will still be lower resolution, and likely to overlap.

We're also going to lose a few features provided by the `LocationMarkers` component:
- The geometry for the currently selected record will no longer be highlighted a different color (#292)
- For projects that use coordinates instead of polygon geometries, we'll no longer be animating the marker for the currently selected record with a "pulsing" animation
- When selecting a record, we'll no longer be fitting the map bounds to the geometries of the selected record

I'm sure there are other caveats to this approach that I'm missing by just looking at the code, so we'll want to thoroughly test with different datasets.

PR #325, opened 2025-11-24
Record detail redesign

camdendotlol:
# Summary

This PR makes a bunch of changes to bring the record detail page closer to the JUEL designs.

* Thicker banner at the top with a background color pulled from the `primary_color` branding value
* The `GeoHeader` component now only shows top-level geometry, i.e. it only shows up when viewing a place record with geometry on the record itself
  * This means that we can refactor the related places fetching to live inside `RelatedPlaces` now that `GeoHeader` doesn't need that data
* `RelatedPlaces` now shows just one column of names and renders a map on the righthand side containing the geometry data for the places within that specific relationship
* The record detail page now uses `text-body` and `text-header` where appropriate (text branding config was previously completely ignored on this page)
* A bunch of small changes to font sizes, spacing, etc.

## Screenshots

(These were taken on a project with the primary color set to black.)

### With cover and map
Screenshot 2025-11-25 at 9 48 26 AM

## With map
Screenshot 2025-11-25 at 9 49 08 AM

## No cover or map
Screenshot 2025-11-25 at 9 52 32 AM

### DIFF
--- a/src/apps/pages/RecordDetail/GeoHeader.astro
+++ b/src/apps/pages/RecordDetail/GeoHeader.astro
@@ -1,13 +1,10 @@
 ---
-import UserDefined from '@apps/pages/RecordDetail/UserDefined.astro';
-import Identifiers from '@apps/pages/RecordDetail/Identifiers.astro';
 import PlaceMap from '@apps/pages/RecordDetail/PlaceMap';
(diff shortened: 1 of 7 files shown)


dleadbetter (commented):
Looks good. Could you add a couple of screenshots that show the changes?

PR #327, opened 2025-11-25
Adding search bar component to Hero

ajolipa:
### In this PR
Adds a `SearchBar` MDX component for use inside rich text components, e.g. in page heroes.
image

### Questions and Notes
- How configurable will we need things like the button color to be?
- I think ideally the `SearchBar` component would receive an `onSubmit` prop rather than forcing the submit action to be redirecting to a URL with the query string appended; however, with the way the data comes in from Tina, and with the way that Astro handles MDX components inside an Astro component it was much easier to just have a script add the `submit` method rather than trying to use React hooks. Open to suggestions on configuring that differently though.


src/apps/pages/sections/mdx/SearchBar.tsx
```diff
@@ -0,0 +1,37 @@
+import { Icon } from '@performant-software/core-data/ssr';
+
+interface SearchBarProps {
+  button_text?: string;
+  placeholder?: string;
+  url: string;
+}
+
+const SearchBar = (props: SearchBarProps) => {
+  const { button_text, placeholder, url } = props;
```
dleadbetter: Could we rename this to camel case?

`const { button_text: buttonText } ....`

src/apps/pages/sections/FreeText.astro
```diff
@@ -9,5 +10,21 @@ const { content } = Astro.props;
 >
   <TinaMarkdown
     content={content}
+    components={richTextComponents}
   />
-</article>
\ No newline at end of file
+</article>
+
+<script>
```
dleadbetter: This seems like an awkward place for this logic. The FreeText component isn't guaranteed to have any search bars and really shouldn't have any knowledge of the MDX child components. Is there any reason we wouldn't want this in the `SearchBar` component?

src/apps/pages/sections/FreeText.astro
```diff
@@ -9,5 +10,21 @@ const { content } = Astro.props;
 >
   <TinaMarkdown
     content={content}
+    components={richTextComponents}
```
dleadbetter: Interesting pull this out into a utility class. Could you explain a little about why we didn't just inline the component?

src/apps/pages/sections/mdx/SearchBar.tsx
```diff
@@ -0,0 +1,37 @@
+import { Icon } from '@performant-software/core-data/ssr';
```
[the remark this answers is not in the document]
ajolipa: I guess I just envisioned keeping the MDX components used on the custom pages all together with the section components, but I'm not attached to that.

src/apps/pages/sections/FreeText.astro
dleadbetter: I think if we're having the component be interactive on the client, we're going to have to allow for some client rendering. Otherwise we'll end up with Javascript business logic disconnected from the presentation, which is going to make the code hard to follow. It would be much cleaner to use React's state management to handle the input value and define a function for `onSubmit`. I think you should be able to use the `client:load` directive to still render the component on the server and have it hydrate on the client.

ajolipa: So you mean option 3 from the list above? Make the whole `free_text` section (and anywhere else we might use the search bar) `client:load`?

tina/content/pages.ts
```diff
@@ -29,6 +29,25 @@ const SpacerSizes = [{
...
+    type: 'string'
+  }, {
+    name: 'button_text',
+    label: 'Button Text',
+    type: 'string'
+  }, {
+    name: 'url',
+    label: 'Search URL',
+    type: 'string',
+    description: 'Should be the full URL excluding only the query string itself, e.g. `https:://mysite.com/search?index[query]=`.'
```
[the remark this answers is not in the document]
ajolipa: If the format will be consistent for all pages; for the map searches it looks like the format is actually `?q=`, but I noticed on NBU that for the list searches (people, archives) that the format is `?people[query]=` etc. And actually it looks like currently the list search implementation on CDP isn't using URL parameters? So I guess that would need to be changed for this approach to work at all. 🤔

ajolipa:
UPDATE: I've reworked this a bit; I don't think it's necessary or quite makes sense to have it as an MDX component, given how the markdown renderer works and where it shows up in the designs, so I've changed it to a regular React component and added it to the `Hero` section for now. As other content sections are implemented, now that the component exists, it can be added there as well (e.g. I intend for it to be a possible template for the centered text with background component).

PR #328, opened 2025-11-25
Adding centered text with background component

ajolipa:
### In this PR
Adds a highly flexible `CenteredText` section to the page builder, with the following properties and configuration:
- A rich text field that will be displayed in the center of the component, with either light or dark text;
- An optional background image;
- Height can be determined either by the text content (clipping any image overflow) or by the image height;
- Width can either be full screen or contained inside a `Container` element;
- If no image provided, the background color can be selected from `primary, secondary, black, white`.

Note this component can also be used for the centered quotes featured in the JUEL designs, with the quotation icon inserted as an image in the rich text.

### DIFF
--- a/src/apps/pages/sections/Banner.astro
+++ b/src/apps/pages/sections/Banner.astro
@@ -101,7 +101,7 @@ const {
           )}
         </div>
       )}      
-      { content && content.children?.length && (
+      { content && content.children?.length ? (
         <div class={clsx(
           'prose max-w-none py-8 lg:py-16 z-10',
           { '!text-white': color === 'white' },
@@ -112,7 +112,7 @@ const {
             content={content}
           />
         </div> 
-      ) }
+      ) : null }
       { search && ( 
         <SearchBar 
           buttonText={search.button_text}


tina/content/pages.ts
```diff
@@ -18,6 +18,13 @@ const SpacerValues = {
   large: 'large'
 };
 
+const ColorValues = {
+  primary: 'primary',
+  white: 'white',
+  black: 'black',
+  secondary: 'secondary'
+}
```
dleadbetter: Missing semi-colon.

tina/content/pages.ts
```diff
@@ -29,6 +36,20 @@ const SpacerSizes = [{
...
+}, {
+  label: 'Secondary (defaults to off-white)',
+  value: ColorValues.secondary
+}, {
+  label: 'White',
+  value: ColorValues.white
+}, {
+  label: 'Black',
+  value: ColorValues.black
+}]
```
dleadbetter: Missing semi-colon.

ajolipa:
> Is this similar to the existing "Hero" page section?

Similar, yes, but with rich text content instead of the more restricted schema of the hero. As with the multi-column component and `text_block`, I agree that ideally these would be combined into one thing but I didn't initially do it that way because of worry about breaking existing projects.

ajolipa:
UPDATE: I decided this component doesn't need to serve like five different functions and instead can serve two-ish functions: subsuming the existing `Hero` component as well as serving as full-width featured/centered text with background elsewhere on the page. So basically I got rid of the `fullWidth` flag and added in the existing `Hero` props as well as keeping the background color options and rich text field from the `CenteredText` component.

I've played with this a bunch in various configurations and I think this feels like the most natural way to combine things, but of course I'm open to suggestions! Certainly the `Banner` component itself is a bit of a tangle of conditionals, but I think the Tina interface for it is pretty clean and it does what we want (namely reproduce the existing look of the heroes, and add the option for rich text content).

src/apps/pages/sections/Banner.astro
```diff
@@ -0,0 +1,107 @@
...
+  )} /> }
+  { darken && <div
+    class='absolute top-0 left-0 w-full h-full bg-gray-900/30'
+  /> }
+  <div class={clsx(
+    { 'absolute inset-0 w-full h-full flex flex-col justify-center': !clip },
+    { 'items-center': textAlignment === 'center' }
+  )} >
+    <Container className={clsx(
+        'flex flex-col justify-center',
```
dleadbetter: There's some wacky white space going on here.

dleadbetter (commented):
This looks great. Just a couple of formatting questions.

src/apps/pages/sections/Banner.astro
ajolipa: Lol so there is. Fixed now!

PR #331, opened 2025-11-25
Adding color options to spacer and adding spacer as MDX component

ajolipa:
### In this PR
Pretty much what it says on the tin: Adds the ability to specify a color from among `primary, secondary, white, black` in the `Spacer` component to display a `2px` horizontal bar of the chosen color. Additionally, an MDX component version of the component is also added to the rich text editor of the `FreeText` field, for further flexibility.

### Notes
It is annoying to me that I had to create a TSX version of the same component, but I got yelled at by Astro when I tried to pass an Astro component to `TinaMarkdown`. I think possibly @camdendotlol once came up with a sneaky way around that, or at least around a similar problem? 

Once this is merged, the rich text template for column content in PR #329 should also have the `Spacer` MDX component added.


src/apps/pages/sections/Spacer.astro
```diff
@@ -1,12 +1,18 @@
 ---
```
dleadbetter: Eek! I don't like have two versions of this same component. Could we add the `.tsx` version to `/src/components` and import it here?

src/apps/pages/sections/mdx/Spacer.tsx
```diff
@@ -0,0 +1,24 @@
+import clsx from "clsx";
```
dleadbetter: Single quotes.

tina/content/pages.ts
```diff
@@ -18,6 +18,13 @@ const SpacerValues = {
   large: 'large'
 };
 
+const colorValues = {
+  black: 'black',
+  white: 'white',
+  primary: 'primary',
+  secondary: 'secondary'
+}
```
dleadbetter: Missing semi-colon.

PR #332, opened 2025-12-02
Gallery page

camdendotlol:
# Summary

This PR adds a new auto-generated gallery page.

The only config item added is `gallery`, which takes a link to the manifest of a Core Data gallery record. Every relationship from the top-level manifest is rendered as an item on the gallery page. Clicking on a gallery opens it in a totally out-of-the-box Clover IIIF viewer.

The "View All Images" button links to the image search page.

## Breaking changes

Clover can't handle relative manifest URLs (e.g. `/api/myManifest.json`) so this PR introduces a required `PUBLIC_BASE_URL` environment variable. This will only be used when constructing the content collection manifest links on a static site.

## Static sites

There's some special handling for static sites to prevent the website from depending on Core Data:

1. During the build process, each child manifest is downloaded and added to a content collection using its URL as ID.
2. A new CDP API route serves these manifests at `/api/gallery/[:id].json` where `[:id]` is a URI-encoded version of the URL from step 1. There really wasn't a better identifier to use here :/

## Screenshot

(the white border on the sides of the scanned document images is part of the images)

Screenshot 2025-12-02 at 12 23 07 PM

### DIFF
--- a/src/utils/galleries.ts
+++ b/src/utils/galleries.ts
@@ -13,11 +13,14 @@ export const fetchJson = async (url: string) => {
 };
 
 /**
- * The project_ids param disappears from the manifest ID when accessed directly from Core Data,
- * so we need to truncate it to match the ID in the content collection.
+ * Generates a URL from a manifest ID with search params removed.
  * @param manifestId
  */
-export const truncateManifestId = (manifestId: string) => manifestId.split('?project_ids')[0];
+export const truncateManifestId = (manifestId: string) => {
+  const url = new URL(manifestId);
+  url.search = ''
+  return url.href;
+}
 
 
 /**
--- a/src/apps/gallery/index.astro
+++ b/src/apps/gallery/index.astro
@@ -30,7 +30,7 @@ const items = manifest.items.map((item) => ({
 
 const { t } = await getTranslations(Astro.currentLocale);
 
-const imageSearchConfig = config.search.find(item => item.type === '/image');
+const imageSearchConfig = config.search.find(item => item.type === 'image');
 ---
 <Container className='my-6'>
   <div class='flex items-center justify-between gap-4'>


src/utils/galleries.ts
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * Gets the list of manifests for the gallery.
+ */
+export const getManifests = async () => {
+  const data = [];
+
+  const json = await fetchJson(config.gallery);
+  const items = json.items || []
+
+  for await (const item of items) {
```
dleadbetter: Do we need `for await` here? Is `items` asynchronous?

src/utils/galleries.ts
```diff
@@ -0,0 +1,74 @@
...
+};
+
+/**
+ * Gets one manifest by its ID.
+ */
+export const getManifest = async (url: string) => {
+  return await fetchJson(url);
+};
+
+export const getStaticManifestEndpointPaths = async () => {
```
dleadbetter: If this isn't used in multiple places, it can probably be done inline in the API endpoint.

src/utils/galleries.ts
```diff
@@ -0,0 +1,74 @@
+import config from '@config';
+import { hasContentCollection } from '@root/src/content.config';
+
+const BASE_URL = import.meta.env.PUBLIC_BASE_URL;
```
dleadbetter: Let's add this to `.env.example`.

src/utils/galleries.ts
```diff
@@ -0,0 +1,74 @@
...
+  const response = await fetch(url);
+  return await response.json();
+};
+
+/**
+ * The project_ids param disappears from the manifest ID when accessed directly from Core Data,
+ * so we need to truncate it to match the ID in the content collection.
+ * @param manifestId
+ */
+export const truncateManifestId = (manifestId: string) => manifestId.split('?project_ids')[0];
```
dleadbetter: Could you explain a little more about what we're doing here? My hope was that the `gallery` config option could be used for any manifest, not just ones coming from Core Data.

src/apps/pages/Gallery/index.astro
```diff
@@ -0,0 +1,30 @@
...
+
+const hasImageSearch = config.search.some(item => item.route === '/media');
+---
+<Container className='my-6'>
+  <div class='flex items-center justify-between gap-4'>
+    <h1 class='text-3xl font-header'>{t('imageGalleries')}</h1>
+    {hasImageSearch && (
+      <a
+        class='bg-primary py-3 px-5 rounded-lg text-white text-sm'
+        href={`/${Astro.currentLocale}/search/media`}
```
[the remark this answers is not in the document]
camdendotlol: #323 added image search. It always uses the list search UI, which gets all its data directly from the TS response and doesn't use the service classes.

PR #333, opened 2025-12-02
Adding multicolumn page builder layout component

ajolipa:
Recreation of PR #329 with a fresh branch.


src/apps/pages/sections/MultiColumn.astro
```diff
@@ -0,0 +1,70 @@
...
+    class={clsx(
+      'w-full grid grid-cols-1',
+      gridCols,   
+      { 'gap-16': gap === 'large' },
+      { 'gap-8': gap === 'small' || !gap }
+    )}
+  >
+    {
+      _.map(content, (col: any) => ( 
+        <Column items={col.content} classNames={total === 12 && col.width} />
```
dleadbetter: Took me a second to figure out what was going on with the `classNames` prop. Could we:
- Use `clsx` to apply conditional class names
- For components, we typically use `className` (singular) as the prop name

src/components/pageBuilder/ColumnContent.astro
```diff
@@ -0,0 +1,37 @@
+---
```
dleadbetter: I think it's fine for the `Column` and `ColumnContent` components to live in either `src/components` or `src/apps/pages`. If the components will only ever be used in the `MultiColumn` section, `src/apps/pages` probably makes the most sense.

src/apps/pages/sections/MultiColumn.astro
ajolipa: Ah sure, I thought I remembered some weird interaction where using `className` as a prop could lead to weird Tailwind behavior where it got confused with the native JSX prop or whatever, but this is probably not a complicated enough use case for that to come up (and/or I'm remembering wrong).

src/components/pageBuilder/ColumnContent.astro
ajolipa: Yeah I wasn't sure. My goal is for things to be as modular and flexible going forward to build on top of as possible; I expect we'll end up reusing `Column` in other page builder components, at the very least, like in the tabbed content section I'm working on now (although that might end up just basically wrapping around the `MultiColumn` component...). My first instinct was to keep the page builder-related components in `src/apps/pages/sections`, maybe in their own folder, but I'm basically agnostic and will put them wherever you think makes sense!

dleadbetter (approved):

PR #339, opened 2025-12-05
Adding link button as MDX component in page builder

ajolipa:
### In this PR
- Adds the `LinkButton` component as an MDX insert in the rich text components inside the page builder;
- Expands the `LinkButton` component to accept specified background and text color and optional arrow icon;
- Refactors the MDX components into a `RichText.astro` component in the page builder, so that when additional MDX components are added they only need to be added in one place.


src/components/LinkButton.tsx
```diff
@@ -1,21 +1,39 @@
 import clsx from 'clsx';
+import { ColorValues } from '@root/tina/content/pages';
+import { ArrowRightIcon } from '@heroicons/react/24/outline';
```
dleadbetter: There's a bit of an inconsistency throughout the code base where some places we're using `@heroicons` and some places we're using icons from `@performant-software/core-data`. Should we make a push to use `@performant-software/core-data` icons throughout and phase out the `@heroicons`?

ajolipa: Oh yeah great catch. Am I missing something or is there currently not a `right_arrow` icon in [our library](https://github.com/performant-software/react-components/blob/master/packages/core-data/src/components/Icon.js)? If so then we should definitely add one and then update this; seems like that should probably be a separate future PR? Or should I try to get that done and incorporated here?

dleadbetter: Ah, you're right. Sure, let's add a separate issue for that.

Issue #311, opened 2025-11-14
Page Builder: Tabbed content


ajolipa:
Allow pages to contain distinct tabs each with their own content.

<img width="1941" height="849" alt="Image" src="https://github.com/user-attachments/assets/3456716f-d8d2-4b40-86bf-d904f2ea19eb" />

jamiefolsom:
Closed by #340.

PR #340, opened 2025-12-09
Adding tab component

ajolipa:
### In this PR
Adds the tabbed content component as described in Issue #311 .

### Notes and questions
- After wrangling a bit with trying to make this component serve both as section (e.g. for displaying team members) and as page-level tabs (where the tabs are incorporated at the bottom of the hero and the whole page content is different for each tab) I decided it makes more sense to just separate those two functions. So this is only the former, meant to be displayed as a section in the page body inside a `Container`.
- Initially I was thinking that the templates for the content inside each tab could be any other section template (or at least the ones that are designed to be inside a `container`), but I held off on that to avoid unnecessarily complication; certainly for the use case in the JUEL designs having the `multi_column` template is sufficient. But I'm open to other suggestions.


src/apps/pages/sections/Tabs.astro
```diff
@@ -0,0 +1,87 @@
...
+            gap={section.gap}
+          />
+        ))
+      }
+    </div>
+  ))
+}
+
+<script is:inline>
+  window.document.addEventListener('DOMContentLoaded', () => {
```
dleadbetter: What are you thoughts on using a React component with `client:load` vs. inline Javascript? Is the performance savings worth the readability/maintainability penalty?

PR #341, opened 2025-12-09
Adding a more full set of theme colors to schema

ajolipa:
### In this PR
Per discussions with Chelsea, adds a more complete set of theme colors to the `branding` configuration and to the color options for various parts of the page builder components.

### Notes
This made my head hurt more than I expected it to, basically because there are two separate uses of the theme colors:
- In the auto-generated site pages (search, record detail) we're using the primary and secondary colors in various places -- header, footer, some of the buttons or pills in the search interface, etc. In those cases we need to programmatically decide whether to use light or dark text on those elements, for accessibility (and just to look good), so we're configuring `textPrimary` and `textSecondary` colors to mean "text color to be used on primary/secondary backgrounds".
- For the page builder components, the user or dev who's creating the pages will have control over the background and text colors, so we're not (I think) trying to programmatically determine the text color based on the background. The designs have a default dark text color, default light text color, and an alternate/secondary text color (in the case of JUEL, a slightly lighter brown than the dark text). So for the purposes of the page builder components, the choices are any of the three theme colors, the default dark text, the default light text, and the alternate text.


ajolipa:
> This brings up a good question: Should we allow the page build to let users select which color to use? Or should the system dictate which parts of the page builder components get which colors? (e.g. the button in the hero component is always "secondary"). The latter would simplify things and reduce the risk of creating inconsistencies between pages.

This is mainly a question for @cblissg I guess; I've been going on the assumption that for the page builder components there should be sort of out-of-the-box defaults but mostly colors should be editable between the different theme colors. But of course I'm open to whatever sounds best to people.

PR #346, opened 2025-12-16
Theming updates

ajolipa:
### In this PR
This PR bundles a bunch of small styling and config updates that were necessary in the course of building out the JUEL main site designs, including...
- Adding the option to make the page header transparent;
- Adding a more complete set of theme colors to the `Branding` config and the CSS setup;
- Creating utility methods for turning Tina field values into text, background, or border color Tailwind classes, for use in many different components;
- Styling various rich text elements (such as headers) according to the designs;
- A few other random little things.

### DIFF
--- a/src/components/Card.astro
+++ b/src/components/Card.astro
@@ -1,6 +1,6 @@
 ---
-import clsx from "clsx";
-import { toBorderClass } from "@utils/pageBuilder";
+import clsx from 'clsx';
+import { toBorderClass } from '@utils/pageBuilder';
 
 const { alt, border, date, dateOptions = {}, category, title, author, slug, imageUrl, blurb, t, rounded } = Astro.props;
 


dleadbetter:
Is there an issue we can think this PR to?

src/components/Card.astro
```diff
@@ -1,11 +1,18 @@
 ---
-const { alt, date, dateOptions = {}, category, title, author, slug, imageUrl, blurb, t } = Astro.props;
+import clsx from "clsx";
```
dleadbetter: Single quotes.

src/styles/index.css
```diff
@@ -72,6 +81,71 @@ h1, h2, h3, h4, h5, h6 {
   color: var(--color-text-secondary);
 }
 
+
+/*
+* Page builder styles
+*/
+
+.page-builder .bg-layout-alt, .page-builder .bg-layout-alt div {
```
dleadbetter: Do we want these styles scoped to just custom pages? I would think we want the branding applied globally.

src/layouts/Layout.astro
```diff
@@ -132,11 +147,20 @@ if (config.gallery) {
...
         customColorSecondary: colorSecondary,
-        customTextPrimary: textPrimary,
-        customTextSecondary: textSecondary,
-        customColorLayout: colorLayout
+        customColorTertiary: colorTertiary,
+        customPrimaryContent: primaryContent,
+        customSecondaryContent: secondaryContent,
+        customTertiaryContent: tertiaryContent,
+        customContent: layoutContent,
+        customContentAlternate: layoutContentAlternate,
```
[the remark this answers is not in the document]
ajolipa: Yeah it's tricky... I was attempting to match the language that's used in the Figma designs as much as possible, which is why I switched to `content` over `text`, but I don't have strong feelings about that either way.

The thing about "main text", "main text inverted" etc. is that those are not the same variables as these; they are the options from which these theme variables are selected. For example, imagine a site with the following theme:

Primary: light pink
Main background: dark purple
Alternate background: white
Main content (for light background): black
Main content inverse (for dark background): white

In this case (with the current naming), we'd have `primaryContent = black` (because the primary color is light), `layoutContent = white` (because the main background color is dark), `layoutAlternateContent = black` (because the alternate background color is light). However, if the theme were the same except the main background was light blue, then `layoutContent` would become `black`.

The reason for doing it this way is that in some cases in the designs we want to select the text color just based on whether it's dark or light (for example on these cards -- this isn't based on a theme color, we're just applying a black gradient over the image to make the light text pop:)
image
So in that case we do want to use `contentInverse`, but in most other cases we want to use the text color based on the themed background value, and so whether it's `content` or `contentInverse` will depend on the value of the background. So that's the difference between `content` (always dark), `contentInverse` (always light), and `layoutContent` (depends on value of `colorLayout`).

Sorry for the overly labored explanation, but that's the complexity I'm trying to wrangle here. The user never has to see any of this -- from the perspective of an editor in Tina, they just have to select the text color they want for dark backgrounds and for light backgrounds, and the correct one will be chosen based on the background color for each section.

src/styles/index.css
ajolipa: Yeah good question; I didn't want to make changes to the styling that might affect random things on the existing search/paths/posts pages without first going through and checking how we're using various heading elements. For example right now at Chelsea's suggestion I'm using `h4` for the color highlighted headings that use the secondary (accent) color, which are scattered throughout the designs; but if someone uses `h4` in a random post they don't necessarily want it to be styled that way. So for now I've just been scoping all my modifications to the stylesheet so that they only affect the new page builder stuff, on the assumption that we're happy with the current styling and branding on the existing pages. But if there are things we need to update or that aren't consistent on the existing pages then we should definitely update those. Should that be part of this PR or its own issue? Since I want to get these page builder PRs merged ASAP so that the JUEL custom pages can be put onto dev I've been trying to keep changes scoped to just the page builder stuff when possible so that I'm not introducing things that need regression testing, but you're definitely right that eventually we'd want to go through the rest of the pages as well and ensure consistency.

dleadbetter: If there's time, I think we should check into making these changes across all pages. It will make all of the pages more consistent and DRY up development to not have to repeat these styles.

PR #351, opened 2025-12-22
Add support for custom navbar via Tina

camdendotlol:
# Summary

This PR adds a new content collection called `Navbar` that can be configured in Tina.

Navbar items *must* include a label. Top-level items *can* contain either a URL or a list of `options`, which is a child array of items. Nested navbar items cannot contain children of their own. If a top-level navbar item contains any children (`options`), it is rendered as a dropdown.

The collection is scoped per locale the same way the i18n files are, e.g. `/navbar/en.json`. If navbar data for a given locale doesn't exist, CDP falls back to the original default navbar logic (this will likely be the case on most projects).

## Note

I think this is my first time working with the Tina side of the codebase, so there may be some boilerplate or extra handling missing here.


dleadbetter:
Is there an issue we can link this PR to?

dleadbetter:
This looks great. Much less code than I was expected 👍 .

> If you provide both a URL and a list of options, the button will be rendered as a dropdown and won't link the URL. It would be nice if we could hide or disable the URL field in Tina in that case, but I couldn't find any obvious/quick way to do that. Suggestions here are appreciated.

It looks like we're overloading the "Navigation Item" object to be either a top level nav item, or a menu of nav items. Could we use Tina's [template](https://tina.io/docs/reference/types/object) concept to break them out into two distinct options?

dleadbetter (approved):
Looks great!

PR #353, opened 2025-12-31
CDP #352 - Accessibility tests

dleadbetter:
This pull request adds some (rudimentary) automated a11y tests using [Playwright](https://playwright.dev/) for all accessible pages in a particular deployment. Tests can be run using `npm run playwright` and will run accessibility tests against Chrome, Firefox, and Safari browsers. In order for the tests to run, the `A11Y_HOST` environment variable must be provided.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
     "start": "node scripts/build.mjs && tinacms dev -c \"astro dev\"",
     "build": "node scripts/build.mjs && npm run test-config && tinacms build && astro build",
     "preview": "astro preview",
+    "test-config": "vitest run config.test.ts",
     "vitest": "vitest run",
     "playwright": "npx playwright test"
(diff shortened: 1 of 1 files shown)


src/apps/session/SearchItem.tsx
```diff
@@ -42,51 +43,53 @@ const SearchItem = (props: Props) => {
         <TranslationContext.Provider
           value={{ lang, t }}
         >
-          <MapSearchContextProvider
+          <SearchConfigContextProvider
```
dleadbetter: @camdendotlol - I noticed a bug here. I remember you doing some work around this when implementing the non-map searches, so I just wanted to make sure I'm making the correct changes here.

camdendotlol (approved):

PR #383, opened 2026-01-12
Adding section margin options

ajolipa:
### In this PR
Improves page builder UX by adding the ability to add top and bottom margin directly to each section, rather than requiring a separate `spacer` section every time; this has a few advantages:
- Allows for negative top margin options, which accommodates some of the design requirements where some buttons or images overlap or partially overlap the previous section;
- Greatly cleans up the visual experience of building a page and saves time in building out designs;
- `Spacer` section can still be used in cases where a dividing line or a background color are desired.

The disadvantage is that spacer/margin sizing is now configured at two different points in the code, and the UI with the negative margins could be confusing to some users. I think the first of those can be solved in the future by just factoring out the definition, perhaps into some theme variables even (`---layout-spacer-lg` etc. or similar) and the second can be mitigated through documentation. I think it will be easier to implement those as separate PRs in the future once the current work is merged to `develop`, since it will require going through a bunch of the content files for the existing JUEL pages and updating field values.

This PR also has a few miscellaneous tweaks and fixes that were noticed while working on the new margin schema, and also adds some UI improvements to the Tina schema in various places.


dleadbetter:
Is there an issue we could link to this?

Issue #414, opened 2026-01-16
Header transparency not working and header blocking dropdown


ajolipa:
<img width="768" height="378" alt="Image" src="https://github.com/user-attachments/assets/7368edad-6df5-4aef-a323-fcde3dc42263" />

It seems that the `transparent` field was somehow dropped from the `pages` collection schema, or never successfully added; this causes the nav bar to always have the primary color background, even when not desired.

Additionally, the nav bar is still blocking the dropdown menus (see also Issue #367 ).

Issue #367, opened 2026-01-07
Dropdown menu appears behind navbar


dleadbetter:
On the "Explore" and "Pages" dropdown menus in the navbar, the menu appears behind the navbar element.

<img width="498" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOFNJREE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--b9b09c257d559953eb4300bdd6c53de828e8a879/Screenshot%202026-01-07%20at%207.59.02%E2%80%AFAM.png" alt="Screenshot 2026-01-07 at 7.59.02 AM.png" />

Issue #392, opened 2026-01-13
Contact form


dleadbetter:
Design and implement a way to allow a contact form (using Netlify forms) to be submitted by site visitors.

<img width="769" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMW1XREE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--c7a70b166ab7574c7e635571e34c8cc2c0bbae9e/Screenshot%202026-01-13%20at%209.25.21%E2%80%AFAM.png" alt="Screenshot 2026-01-13 at 9.25.21 AM.png" />

PR #415, opened 2026-01-16
Fixing transparency and z-index of nav bar

ajolipa:
### In this PR
Addresses the issues raised in Issue #414 (and see also Issue #367 ), as well as cleanup of a repeated import.


src/layouts/Header.astro
```diff
@@ -22,7 +22,7 @@ const {
 
 <div
   class={clsx(
-    'flex flex-row justify-center w-full z-30 text-text-primary',
+    'flex flex-row justify-center w-full z-10 text-text-primary',
```
dleadbetter: I ran into this in #392. It's also blocking the notifications panel. Do we need to specify a z-index at all?

ajolipa: With no z-index it ends up hidden behind the hero on pages that have the navigation on top of the hero section.

dleadbetter (approved):

Issue #374, opened 2026-01-07
Footer button padding


dleadbetter:
Left/right padding is not the same on the footer buttons.

<img width="978" height="291" alt="Image" src="https://github.com/user-attachments/assets/6446a37d-f4a9-4506-a4c1-cb861aa0f292" />

<img width="673" height="335" alt="Image" src="https://github.com/user-attachments/assets/4ef67a36-3894-456a-b1da-0774360eeaaf" />

PR #418, opened 2026-01-16
CDP #374 - Footer buttons

dleadbetter:
This pull request fixes the issues with the `LinkButton` components in the footer presented in #374.

**Disclaimer:** This seems to have broken some of the places where this component is used as a link without padding and with underline text decoration. I would argue that the purpose of this component is to format a `<a>` tag as a button, so perhaps we need a separate component for when we want to render a more traditional link?

Screenshot 2026-01-16 at 6 49 34 AM
Screenshot 2026-01-16 at 6 50 07 AM

### DIFF
--- a/src/components/LinkButton.tsx
+++ b/src/components/LinkButton.tsx
@@ -9,33 +9,47 @@ interface Props {
   color?: string;
   content: JSX.Element | string;
   href: string;
+  target?: string;
   text?: string;
 }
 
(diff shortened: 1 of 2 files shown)


camdendotlol (approved):
Looks good.

Going by https://github.com/search?q=repo%3Aperformant-software%2Fcore-data-places%20linkbutton&type=code this component doesn't seem to be used in a lot of places so the issue in the disclaimer should be easy to take care of.

PR #433, opened 2026-01-22
Adding published flag and post categories

ajolipa:
### In this PR
- Adds a `publish` flag to the `posts` model to optionally use a draft/published workflow where posts aren't shown on the frontend until they're marked `publish`;
- Adds a `categories` list to the posts schema which is populated from a list of values set in the `config.json` file; when present it will add a dropdown menu to the posts page where viewers can select a category to filter by.
image

### Notes
- I was initially going to add this all to `paths` as well because why not, but it looks like paths hasn't yet been updated to the progressing loading thing, so it would be a few more steps to add the filtering and category stuff there.
- I'm using the HeadlessUI `ListBox` component for the dropdown since as best I could tell we don't have a simple list select component built out in our `react-components` library, but it's definitely possible that I just missed something. In the future if we do add such a component to the library, this should probably be updated to use it.

### DIFF
--- a/src/apps/posts/PostList.tsx
+++ b/src/apps/posts/PostList.tsx
@@ -19,7 +19,6 @@ const PostList = (props: Props) => {
   const { lang } = props;
 
   const [cursor, setCursor] = useState<string | null>(null);
-  const [showMore, setShowMore] = useState<boolean>(false);
   const [posts, setPosts] = useState<any[]>([]);
   const [category, setCategory] = useState<string | null>(null);
   const [loading, setLoading] = useState<boolean>(false);
@@ -41,8 +40,7 @@ const PostList = (props: Props) => {
     const res = await fetchPosts(params);
 
     setPosts(prev => prev.concat(res.posts));
-    setCursor(res.metadata?.endCursor);
-    setShowMore(res.metadata?.hasNextPage);
+    setCursor(res.metadata?.hasNextPage ? res.metadata?.cursor : null);
     setLoading(false);
   }, [category]);
 
@@ -98,7 +96,7 @@ const PostList = (props: Props) => {
           </a>
         )) }
       </div>
-      { showMore && !loading && (
+      { cursor && !loading && (
         <div className='w-full flex justify-center items-center py-6'>
           <Button
             onClick={() => onLoadPosts(cursor)}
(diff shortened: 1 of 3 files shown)


src/apps/posts/PostList.tsx
```diff
@@ -27,39 +34,71 @@ const PostList = (props: Props) => {
...
+        }
       </div>
       <div className='flex flex-col italic text-lg divide-y divide-secondary text-[#222222]'>
+        { loading && <p className='not-italic text-base'>{t('loading') || 'Loading'}...</p>}
         { _.map(posts, (post) => (
           <a
             className='hover:underline hover:font-semibold py-4 font-header'
-            href={getRelativeLocaleUrl(lang, `posts/${post._sys.filename}`)}
-            key={post._sys.filename}
+            href={getRelativeLocaleUrl(lang, `posts/${post?._sys?.filename}`)}
```
dleadbetter: Interesting. Will the `posts` array ever contain null/undefined values?

tina/content/settings.ts
```diff
@@ -21,6 +21,34 @@ const Settings: Collection = {
       name: 'localize_pages',
       label: 'Localize Pages',
       type: 'boolean'
+    }, {
+      name: 'posts_config',
```
dleadbetter: Let's document these changes in the README.

src/apps/posts/PostList.tsx
```diff
@@ -27,39 +34,71 @@ const PostList = (props: Props) => {
...
+                    </ListboxOption>
+                  )) }
+                </ListboxOptions>
+              </Listbox>
+            </div>
+          )
+        }
       </div>
       <div className='flex flex-col italic text-lg divide-y divide-secondary text-[#222222]'>
+        { loading && <p className='not-italic text-base'>{t('loading') || 'Loading'}...</p>}
```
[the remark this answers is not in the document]
ajolipa: Oh yeah good catch, this was an impatient step to test it out quickly which I never went back and fixed. Thanks!

tina/content/settings.ts
dleadbetter: Should have also mentioned, can we add a test for this in `config.test`?

dleadbetter (approved):

PR #460, opened 2026-01-27
Add `range` as accepted facet type in config

### LINKED ISSUE #438: Allow `range` type for facets
Right now in the config if you specify type `range` for a facet, the build will fail at the testing stage because the tests look for the type to be either `list` or `select`. I assume that's to stop people from accidentally specifying that a non-numerical field should be `range`, but we should be able to handle that case and still allow for numerical facets to be rendered as range sliders.

ajolipa:
### In this PR
Resolves #438 by adding `range` to the set of allowed `type` values for items in the `search.facets` field in the config. This is necessary because in some data models (e.g. JUEL) we may have numeric fields on models that aren't the dates of associated events, which is currently the only thing we display as a range facet by default I believe.

Note that upon testing it seems that a non-numerical field with `range` as the specified type will simply not render, and will not crash the page or cause any errors, so the danger is low in giving users this power.


dleadbetter:
How much longer do you think it would take to fix in `react-components`? I think it would be worth it so that users do not have to manually set the type on the facet.

ajolipa:
Closing this since we addressed the issue in `react-components` instead.

PR #494, opened 2026-02-06
Adding EditionCrafter viewer option for detail pages

ajolipa:
### In this PR
Adds the ability to configure an `editioncrafter_config` object in the `detail_pages.<model>` object in the config. If supplied, the `GeoHeader` on the record detail page will be replaced with the EditionCrafter viewer for the corresponding document, as specified via a FairCopy ID field in FairData. Note as currently configured this feature is only available for models whose records are associated with documents in a FairCopy Cloud 2 project.

### Notes and further work
- The back arrow at the top of the detail page was already pretty broken, since the embedded maps use hash routing to track state so you often had to click several times to actually navigate to the previous page (and the text "Back to Items" or whatever is a lie in any case; it's just calling `history.back()`). This attempts to fix the first of those problems by actually tracking the referrer URL when the page first loads and taking you back there. It does not solve the second problem of the wording being misleading.
- This is currently using a beta release of EC that I just made that fixes some janky behavior of the competing hash routes between the embedded map and EditionCrafter; that fix should be further tested and incorporated into whatever the next real release of EC is, and then CDP updated to use the new version.


src/apps/pages/RecordDetail/EditionCrafterViewer.astro
```diff
@@ -0,0 +1,28 @@
...
+const validManifest = await fetch(manifestURL).then((res) => (res.status === 200));
+
+---
+{ validManifest && ( 
+  <div class='w-full pb-6 relative h-[800px]'>
+    <div class='bg-primary h-[80px] w-full' />
+    <div class='absolute gap-6 w-full top-0 h-full'>
+      <Container className='h-full'>
+        <EditionCrafter
+          documentName="Test"
```
dleadbetter: Is this a required prop? I'm not sure we want to hard-code "Test" as the value. Is there any other way to get a name from the current record?

package.json
```diff
@@ -31,6 +31,7 @@
     "@astrojs/sitemap": "^3.4.1",
     "@astrojs/ts-plugin": "^1.10.4",
     "@auth/core": "^0.37.4",
+    "@cu-mkp/editioncrafter": "^1.3.1-beta.12",
```
dleadbetter: I think this is fine for testing (I do it all the time with `react-components`), but will we want an actual release before merging this?

README.md
```diff
@@ -107,35 +107,19 @@ Copy the `/public/config.json` file into your content repository to `/content/se
...
-| detail_pages.models.places.related_manifest.model               | String  | Model to use for the related manifest                                                                                                                                                        |
-| detail_pages.models.places.related_manifest.relationship        | String  | ID of the relationship to display in the viewer                                                                                                                                              |
+| detail_pages                                                    | Object  | Configuration for the detail pages. `<modelname>` can be any of `events`, `instances`, `items`, `organizations`, `people`, `places`, `taxonomies`, or `works`                                |
+| detail_pages.models.<modelname>.editioncrafter_config                 | Object  | Optional configuration for an EditionCrafter viewer (requires associated FairCopy Cloud project)                                                                                       |
```
dleadbetter: We'll want to update the Tina `settings` schema, otherwise this will contribute to issues with editing the config via Tina.

PR #497, opened 2026-02-06
Add missing config fields to Tina

camdendotlol:
# Summary

This PR adds a bunch of config fields that were missing from the Tina interface, including the recently-added `result_card.relationships` array.


dleadbetter (commented):
Looks good. Any idea why `package-lock.json` might be changing?

PR #499, opened 2026-02-10
Adding clover viewer to image search

### LINKED ISSUE #381: Clover viewer in image search
The image search should open a Clover viewer in a similar style to the gallery page when the user clicks on a record.

### LINKED ISSUE #481: Image Search Page Issues
- [ ] Images aren't clickable and don't open to a detailed view - Need to add clover viewer
- [ ] Only Image name should be shown on card and not related items count
- [ ] Height of image is wrong - Need to add h-[340px] . 
```
<img
  alt={props.hit.name}
  className='object-cover w-full h-[340px]'
  src={props.hit.thumbnail}
/>
```

ajolipa:
### In this PR
Resolves #481 (and also resolves #381 ) by adding a clover viewer to the image search page.
image

### Notes
It looks like we're not currently indexing the manifest URL directly in Typesense from what I can tell, but in the case of media records from FairData the manifest URL is determinable from the thumbnail URL, so I'm exploiting that there. If that relationship between the thumbnail and manifest URL were ever broken, this would break. I don't think that's a big deal for now because the whole search interface relies on specific data structures from FairData, but it's worth noting.

### DIFF
--- a/src/components/custom/default/ImageHit.tsx
+++ b/src/components/custom/default/ImageHit.tsx
@@ -1,13 +1,21 @@
 import PropTypes from 'prop-types';
 import { HitComponentProps } from '@types';
+import clsx from 'clsx';
 
 const ImageHit = (props: HitComponentProps) => {
   return (
-    <div className='bg-white rounded-md shadow-xs hover:shadow-md overflow-hidden flex flex-col divide-zinc-200 divide-y h-auto'>
+    <div className={clsx(
+      'bg-white rounded-md shadow-xs hover:shadow-md overflow-hidden flex flex-col divide-zinc-200 divide-y h-auto',
+      { 'cursor-pointer': props.hit.thumbnail }
+    )} onClick={() => {
+      if (props.hit.thumbnail) {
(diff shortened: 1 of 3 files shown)


src/components/custom/default/ImageHit.tsx
```diff
@@ -1,13 +1,21 @@
...
 
 const ImageHit = (props: HitComponentProps) => {
   return (
-    <div className='bg-white rounded-md shadow-xs hover:shadow-md overflow-hidden flex flex-col divide-zinc-200 divide-y h-auto'>
+    <div className={clsx(
+      'bg-white rounded-md shadow-xs hover:shadow-md overflow-hidden flex flex-col divide-zinc-200 divide-y h-auto',
+      { 'cursor-pointer': props.hit.thumbnail }
+    )} onClick={() => {
+      if (props.hit.thumbnail) {
+        props.setManifestUrl(props.hit.thumbnail?.replace('/thumbnail', '/manifest')); //this is a hack since it looks like we're not storing the manifest URL in typesense?
```
dleadbetter: I assume we don't want to spend the time adding the manifest to the Typesense documents?

Issue #294, opened 2025-11-10
Slow page loading in SSR mode


ajolipa:
It can take up to five or six seconds to load content pages (Home, About, etc) or record detail pages. I assume the reason for that is that we're asking Astro to fetch a whole bunch of data (from Tina or from the content API) before building the page at all. Some combination of more intelligent CDN caching and the use of server islands would probably make a difference?

PR #501, opened 2026-02-10
CDP #294 - Performance

dleadbetter:
This pull request makes the following changes in an attempt to speed up page load times:

- Refactors the `Layout` components and turns the `Header` and `Footer` components into Astro server islands
- Updates main content for all pages to load independently from layout as Astro server islands
- Reorganizes the code for rendering pages/detail pages into separate `apps` sub-directories
- Removes unnecessary uses of `useTranslations` hook
- Adds the `DISABLE_CACHE` environment variable to allow disabling the Netlify CDN cache for debugging purposes

#### Notes
- One of the challenges with de-coupling the layout rendering from the data fetching is that we lose the ability to easily pass a value to the `<title></title>` HTML element in the `Layout` component. To get around this, I've implemented the `pages` nanostore which will allow content pages to set the page title based on data fetched from Tina or Core Data. I welcome comments on this strategy, as I'm still unsure how I feel about it.
- Another drawback of the server islands loading independently is that sometimes the main page content will load, then the header will pop in. It can be a little weird looking, but probably worth the increase in performance?


dleadbetter:
I found a `await getTranslations` call in `ColumnContent.astro`, which meant we were going to the Tina API for every column on the page (see [88c9c1c](https://github.com/performant-software/core-data-places/pull/501/commits/88c9c1ca6196c598280f5869d2ca38e065b3772e)). I've moved that call up the tree to avoid unnecessary API calls.

camdendotlol (approved):

Issue #493, opened 2026-02-02
Solutions for Detail Pages


cblissg:
We need solutions for the following missing sections on the detail pages: 

- Video Animations
- Panoramas
- Essays
- Thumbnails of Rendering and Historical Images
- Second design option for displaying related places (Academical Village vs Places)

Goal for launch is to have at least the Thumbnails and a way for them to access animation, panoramas, and essays. 

**Solution for Thumbnails:** 
[Clover IIIF Slider
](https://samvera-labs.github.io/clover-iiif/docs/slider)
<img width="377" height="154" alt="Image" src="https://github.com/user-attachments/assets/3e0b8f0d-6ab4-48a4-a32a-8eae39005b49" />

**Temporary solution for animations, panoramas, and essays:**
New model for links that can be related to people/places

<img width="844" height="155" alt="Image" src="https://github.com/user-attachments/assets/662c972c-bd9d-4fd0-a2f2-30aaded55630" />

PR #510, opened 2026-02-12
Update related media on detail pages to use embedded clover viewer

ajolipa:
### In this PR
Addresses part of Issue #493 by updating the `RelatedMedia` component to directly embed a clover viewer in the page, rather than using small thumbnails and a modal viewer. The viewer accepts the full collection manifest for the record, meaning that users can select from a dropdown to see different relationships between the models (in the case of the JUEL sites, "Historical Images" and "Renderings").

image


ajolipa:
> This will change the way that related media records are presented for all sites. Is that the idea?

I believe this is the idea! Tagging @cblissg for confirmation.

Issue #518, opened 2026-02-17
Detail page back button


dleadbetter:
The back button on the detail page is not longer working.

jamiefolsom:
Closed by #522.

PR #522, opened 2026-02-17
CDP #518 - Back button

dleadbetter:
This pull request updates the record detail page back button to use `history.back()` to navigate back to the previous page. Since we've moved the page content to a server island, document.referrer will no longer provide the URL of the previous page.

I know this was changed recently due to issues with the button not working in all cases. I was unable to reproduce the problem using `history.back()`, but will certainly look into it if it can be reproduced.

### DIFF
--- a/src/apps/detailPages/RecordDetail.astro
+++ b/src/apps/detailPages/RecordDetail.astro
@@ -53,8 +53,8 @@ const faircopyId = ecViewer && record.user_defined ? record.user_defined[ecViewe
   <Container>
     <div class='pb-6 flex items-center gap-2'>
       <BackButton
-        id='backButton'
         size={24}
+        client:load
       />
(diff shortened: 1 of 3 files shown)


ajolipa:
> This pull request updates the record detail page back button to use `history.back()` to navigate back to the previous page. Since we've moved the page content to a server island, document.referrer will no longer provide the URL of the previous page.
> 
> I know this was changed recently due to issues with the button not working in all cases. I was unable to reproduce the problem using `history.back()`, but will certainly look into it if it can be reproduced.

The issue with `history.back()` is that when you have hash navigation like EditionCrafter uses, the button will just take you back one step in that navigation, rather than back to the separate page you came from. See e.g. https://deploy-preview-522--juel-staging.netlify.app/en/items/e2ffffa7-9155-47d3-8663-c3b7590cbfc8 .

camdendotlol (approved):

PR #524, opened 2026-02-17
Add Clover media embed to pages

camdendotlol:
# Summary

This PR adds a Clover embed to pages identical to the one used in posts, including some refactoring to share code in Tina and on the frontend.

## Note

I think this was my first time doing anything significant in the Tina post/page code, so I might have missed something, but I wonder if we could share more of the embeds between post and page this way.


camdendotlol:
The failing preview deployments seem to be from the ongoing Netlify issues. It might be better to wait a day and make sure though.

dleadbetter (commented):
Code looks good. Any thoughts on the build errors?

PR #571, opened 2026-03-02
Adding sensitive content warning/filter to detail and search

ajolipa:
### In this PR
This PR does two things:
- On image searches, adds a toggle to the search facets list to hide results with `content_warning: true`;
- On record detail pages, adds a check for whether any of the related media have a content warning, and if so hides the media viewer behind a button the user has to click to proceed.

### Notes
- Should the search filter always appear for any image search, or should that be configurable?
- These features rely on the two issues described here being addressed: https://github.com/performant-software/core-data-cloud/issues/557

### DIFF
--- a/src/apps/detailPages/RelatedMedia.astro
+++ b/src/apps/detailPages/RelatedMedia.astro
@@ -1,6 +1,7 @@
 ---
 import MediaContents from '@apps/detailPages/MediaContents';
 import ServiceFactory from '@services/coreData/factory';
+import _ from 'underscore';
 
 interface Props {
   model: string;
@@ -11,6 +12,9 @@ const { model, uuid } = Astro.props;
 
 const service = ServiceFactory.getService(model);
 const manifests = await service.getRelatedManifests(uuid);
+const { mediaContents } = await service.getRelatedMedia(uuid);
+const showContentWarning = _.some(mediaContents, (img) => (img.content_warning));
+
 ---
 {manifests && manifests?.items?.length > 0 && (
   <MediaContents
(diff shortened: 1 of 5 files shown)


src/apps/search/list/SensitiveContentToggle.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { useContext } from "react";
+import { useSearchConfig } from "../SearchConfigContext";
+import { useToggleRefinement } from "react-instantsearch";
+
+const SensitiveContentToggle = () => {
+  const { t } = useContext(TranslationContext);
+  const config = useSearchConfig();
+  
+
+  if (!(config.type === 'image')) {
```
camdendotlol, line 12: Let's do `config.type !== 'image'` for clarity

src/apps/detailPages/RelatedMedia.astro
```diff
@@ -11,12 +12,16 @@ const { model, uuid } = Astro.props;
 
 const service = ServiceFactory.getService(model);
 const manifests = await service.getRelatedManifests(uuid);
+const { mediaContents } = await service.getRelatedMedia(uuid);
```
camdendotlol, line 15: The content warning field should be included in the manifests, so it would be better for load times to get it from there instead of adding a second fetch request.

See https://github.com/performant-software/core-data-places/blob/b50eddcb83e5cb3cf301152448c16764e7199dda/src/apps/gallery/ManifestGrid.tsx#L29 for an example that can probably be factored out into a helper function and shared between this component and that one.

camdendotlol (changes requested):
Looks good, just a couple small changes.

PR #572, opened 2026-03-03
Paths and Posts visual editing

ajolipa:
### In this PR
Adds visual editing capabilities for the `posts` and `paths` models.

https://github.com/user-attachments/assets/62e00f30-154f-4149-9034-4196cd3491ca

Note that the page loaded into the visual editor is a `preview` URL and can thus be treated differently from the posts/paths as displayed on the main website (for example, not filtered by publication status). Each preview lives at a link that includes a simple hash created from the post name together with an environment variable. This is not secure as such (in order to work with the Tina UI, the environment variable had to be accessible on the client side) but should suffice to make the preview links difficult to stumble upon without significant effort, which seems sufficient in this case for the goal of not having public access to unpublished content.


blms (approved):
Looks good! I'm totally fine with the security of the preview URL approach. Tested locally a tiny bit and visual editing seemed to work well.

PR #577, opened 2026-03-18
Initial pass at adding clerk login flow to tina

ajolipa:
### In this PR
Migrates SSO from Keycloak to Clerk.
image

### Notes and limitations
- To make the Clerk integration work smoothly, TinaCMS was updated to the latest version (3.6.1), along with some of its related packages. The Keycloak integration no longer works as previously configured in the older version, so is removed entirely in this PR. If we wanted to support both Keycloak and Clerk simultaneously (which I don't think we do), more work would have to be done to figure out how to make that possible.
- Ideally we would hide or disable any fields that the current user is not allowed to alter, but that would involve creating custom components for each of them that work along the same model as the current `PublishToggle` component. Certainly possible, but a lot of busywork (maybe this is where Claude could actually step up and help?).
- If a non-admin user attempts to delete or rename a post from the list of all posts, the whole page crashes, due to some sort of bug on Tina's end about how the error handling is done. If you reload the page everything's fine and the document was not deleted/renamed.

### How to test
- [x] Go to https://padp.staging.performant.studio/admin/ and attempt to log in with an account that's a member of the Performant Software organization within the Performant Studio Staging application. You should receive an email code (or enter your password, depending on the user settings in Clerk I believe) and successfully log in to Tina. Once logged in you should see the PADP staging homepage and the Tina sidebar on the left.
- [x] Click on the hamburger menu in the top left and go to the Posts collection. You should be able to create a new post. When you create your post, you should not be able to edit the Published field. (You should just see the word "Unpublished" and not be able to interact with it.) The post should publish successfully and redirect you once saved to the visual editor view.
- [x] You should now see your own email as the creator, and be able to edit any field except for the creator and published fields, using the visual editor. You should be able to save your edits. 
image

- [x] If you attempt to do anything other than edit your own post, you should receive some sort of error message. (Note: this is the main thing I would like to improve if there's any further time -- the error messages currently are not helpful to the user, and in some cases put the site in a temporarily broken state (fixed by refreshing).) This includes editing other people's posts, attempting to delete or rename a post, or trying to access any of the collections other than Posts and Paths.
- [x] Log out using the button at the bottom of the sidebar menu. 
image

- [x] Log in again, this time with a user with the admin role in the organization. You should now be able to do anything you want, including editing other collections, deleting posts, and toggling the Published state of posts. (Note even admins cannot currently edit the Creator field of a post.) Go ahead and publish the post that your student account made.
- [x] Log out and back in as the student. Now that your post was published, you should no longer be able to make edits to it.
- [x] All of the above should also work for the Paths collection.
- [x] Log out and attempt to log in as a user who is not part of the correct organization. You should not be authenticated.


scripts/build.content.mjs
```diff
@@ -20,6 +20,15 @@ export const fetchContent = async () => {
   // Copy the "content" folder to the current directory
   fs.cpSync(`${TEMP_DIR}/content`, './content', { recursive: true });
 
+  // Append any custom Netlify config to the main one
```
jamiefolsom, line 23: See #594 -- both frames of the Tina Visual Editor have to be served from the same origin, likely the *.performant.studio subdomain, or else we get a cross origin security error.

jamiefolsom (approved):
Staging is working for me as described. There is a short list of other issues to be addressed, but that need not block merging this PR. #602

PR #584, opened 2026-03-23
return 404 for unpublished posts and paths in SSR

jamiefolsom:
## Summary

- Check `published` status in post and path page routes before rendering
- Return 404 for unpublished content in SSR mode

In SSR mode, dynamic routes handle any slug regardless of `getStaticPaths`. The `getStaticPaths` filter (`published: { eq: true }`) only affects static builds. This meant unpublished posts/paths were accessible by direct URL on SSR deploys.

The API routes (`/api/posts/`) already filter by published status, so post/path lists were correct — only direct URL access was affected.

## Test plan

- [ ] Create a post in Tina without checking "published"
- [ ] Visit that post's URL in an incognito window — should get 404
- [ ] Create a post and check "published" — should be accessible
- [ ] Same for paths
- [ ] Post/path list pages should continue to show only published content

### DIFF
--- a/src/apps/paths/Path.astro
+++ b/src/apps/paths/Path.astro
@@ -3,8 +3,12 @@ import PathViewer from '@apps/paths/PathViewer';
 import { fetchPathResponse } from '@backend/tina';
 import PageStoreProvider from '@components/PageStoreProvider';
 
-const { slug } = Astro.props;
+const { slug, publishedOnly } = Astro.props;
 const pathResponse = await fetchPathResponse(slug);
+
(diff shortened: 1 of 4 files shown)


camdendotlol (changes requested):
Currently, we don't fetch the post/path in the Astro component in order to make the initial page load faster, and then the `Post` and `PathViewer` components load the content. Adding the `fetchPost` and `fetchPath` calls in the top-level Astro components would increase initial page load times and duplicates the fetch call that occurs in the child component.

We should implement the filtering in the `Post` and `PathViewer` components, where the data is currently being fetched, instead.

PR #595, opened 2026-03-26
Redeploy button

camdendotlol:
# Summary

This PR adds a new redeploy button to the Tina settings page when running with Clerk authentication. The functionality is provided by an edge function that talks to Netlify's API using a personal access token. Use of this feature will require a `NETLIFY_TOKEN` environment variable to be present.

Some notes on the edge function's behavior:
* it checks that the user is in `TINA_PUBLIC_CLERK_ORG_ID` and is an admin of that org, and will return a 401 error if not
  * this is the same check used by the settings page that the button lives on, so we don't need any extra visibility logic on the frontend
* before triggering a build, the function checks for any currently-running builds. If a build is already running, it returns some helper text instead of triggering another one

## Screenshot

(in this case, there's a message indicating that a build is already running)

Screenshot 2026-03-26 at 3 11 36 PM


ajolipa (approved):
Code looks good, so I'll go ahead and merge this so it can be tested on the `RB-clerk-auth` branch.

PR #607, opened 2026-04-08
Fuzzy places

camdendotlol:
# Summary

This PR adds fuzzy place support to CDP via a new `CertaintyLayer` component.

The feature requires the optional `map.geometry.properties` config value to know where in the Typesense index to look for the `properties` object, similar to how we handle the `geometry` path config.


blms (commented):
Looking good! Could we include a note in `docs/upgrade-notes.md` about the config change?

camdendotlol:
I don't think we need to mention it in the upgrade notes since it's not a breaking change.

blms (approved):
Ah, okay! Makes sense.

PR #608, opened 2026-04-13
Fixing visual editing errors from post inserts

ajolipa:
### In this PR
Fixes an issue where attempting to insert certain MDX components in a path or post crashes the visual editor. The issue seems to be in the component attempting to render on the page before containing any data, so this fix just adds a check for whether any data was actually passed to the component and returns `null` if not.

### DIFF
--- a/src/visualizations/Table.tsx
+++ b/src/visualizations/Table.tsx
@@ -8,24 +8,20 @@ import _ from 'underscore';
 const Table = (props: DataVisualizationProps) => {
   const { t } = useContext(TranslationContext);
 
-  if (!props.data) {
-    return null;
-  }
-
   /**
    * Memo-izes the data as JSON.
    */
-  const { data } = useMemo(() => JSON.parse(props.data), [props.data]);
+  const { data } = useMemo(() => props.data ? JSON.parse(props.data) : null, [props.data]);
 
   /**
    * Memo-izes the table columns and labels.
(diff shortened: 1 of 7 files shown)


src/visualizations/Table.tsx
```diff
@@ -8,6 +8,10 @@ import _ from 'underscore';
 const Table = (props: DataVisualizationProps) => {
   const { t } = useContext(TranslationContext);
 
+  if (!props.data) {
+    return null;
```
camdendotlol: The logic in this PR is good, but [the `return` statements should be below all hooks](https://react.dev/reference/rules/rules-of-hooks), including the `useMemo` that appears under this early return in most of these components.

If `props.data` were to evaluate to something truthy and then become `null` again on the next render, React would freak out at runtime because `useMemo` isn't being called on every render after the first time it's called.

camdendotlol (changes requested):

src/visualizations/Table.tsx
camdendotlol: Alternatively, if the `useMemo` is where the null error is occurring, we could add an `if` statement in there to handle it instead of an early return (up to you).

ajolipa: Good catch, will fix.

camdendotlol (approved):

PR #618, opened 2026-05-01
Saved searches restore an empty map on sites without preloaded geometry

### LINKED ISSUE #616: Saved search from Places view doesn't render map markers when restored
## Summary

A user can save a search from the Places explore view, but when the saved search is restored, the results list populates correctly while the **map shows no markers**. May be related to the fuzzy-places / certainty-radius feature (#607, #612, #614).

## Observed on

- Marronage staging (`deploy/marronage` branch)
- URL: https://marronage.staging.performant.studio/en/search/places/

## Reproduction steps

1. Go to the Places search view
2. Apply any filter (e.g., a Country facet selection) or let the default query run
3. Click "Save search" and give it a name
4. Navigate to Saved Searches and open the saved one

## Expected

The map renders the same set of markers (and certainty radii) that were visible when the search was saved.

## Actual

The map renders the base layer but no place markers or radii. NB: the table and timeline views work, and show data.

## Notes

- Could be timing (map layer tries to read results before they're populated from the session store)
- Could be the new "hide point when certainty radius present" logic on #614 failing to render the radius at all when coming from a restored session
- Console contains 404 errors; see sample saved search. https://marronage-staging.netlify.app/en/sessions/search/53919d62-8ad6-4081-8d27-1abcd6bb9790#map=1/0/0

jamiefolsom:
## What

When a user restores a saved search, the map comes back empty — no place markers, no certainty radii. Affects any deployed CDP site that doesn't preload geometry (`PRELOAD_MAP=true`). List and timeline views work; only the map breaks.

## Why

`SaveButton` saves both `hits` and `features` into the session. On restore, the map sees `features` and tries to fetch each one's geometry from `/api/geometry/[uuid].json`. That endpoint is prerendered from the `geometry` content collection — but the collection is only registered when `PRELOAD_MAP=true`. On sites without the preload flag, every fetch returns 404 and the map renders nothing.

## Fix

- `Map.tsx` — build features directly from `hits`, the same way the live search page does. Fall back to per-feature `fetchGeometry` only when hits are absent (legacy saved sessions on preload-enabled sites).
- `SaveButton.tsx` — stop saving `features`; persist only `hits` and metadata.

Closes #616.

## Testing

- Save a fresh search, restore it, confirm markers render — including certainty radii where present.
- Live `/en/search/places/` and `/en/search/events/` views are unchanged.

### DIFF
--- a/src/apps/search/map/SaveButton.tsx
+++ b/src/apps/search/map/SaveButton.tsx
@@ -1,4 +1,3 @@
-import MapSearchContext from '@apps/search/map/MapSearchContext';
 import { useSearchConfig } from '@apps/search/SearchConfigContext';
 import { saveSession } from '@backend/api/session';
 import TranslationContext from '@contexts/TranslationContext';
@@ -10,7 +9,6 @@ import {
 } from '@performant-software/core-data';
 import NotificationsStore from '@store/notifications';
 import { useCallback, useContext, useState } from 'react';
-import _ from 'underscore';
 
 const SaveButton = () => {
   const [name, setName] = useState<string>();
@@ -20,7 +18,6 @@ const SaveButton = () => {
   const { name: searchName } = useSearchConfig();
   const hits = useCachedHits();
 
-  const { features } = useContext(MapSearchContext);
   const { t } = useContext(TranslationContext);
 
   /**
@@ -48,8 +45,7 @@ const SaveButton = () => {
 
     const data = {
       data: {
-        hits,
-        features: _.filter(features, (f) => f.properties.visible)
+        hits
       },
       name,
       searchName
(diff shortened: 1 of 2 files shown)


jamiefolsom:
Verified locally on `deploy/marronage` with this commit cherry-picked.

Steps: `cdp-dev marronage --skip-clean`, ran a Places search with a Type filter, saved as `test-616-fix`, restored from Saved Searches.

Result: map renders place markers and certainty radii on restore. DevTools Network tab shows **no** `404 /api/geometry/<uuid>.json` calls during restore — confirms `Map.tsx` takes the new hits-first path. Table and Timeline views unchanged. Live `/en/search/places/` and `/en/search/events/` unchanged.

The cherry-pick was reset locally — fix lives only on this branch. Leaving open for dev review before merging to `develop`.

camdendotlol:
I wonder if this is the same issue that https://github.com/performant-software/core-data-places/pull/561 fixed, but this time on saved searches as opposed to embeds. I'll dig deeper into it on Monday.

camdendotlol (approved):
Looks good

PR #622, opened 2026-05-05
Fixing error where Tina config tried to build Clerk instance without config

ajolipa:
### In this PR
It seems that the Netlify builds fail if the Clerk environment variables aren't configured, even though in that case the Clerk auth provider isn't actually used. This should fix that issue.

EDIT: I am not sure why all the deploy previews are still failing; more investigating to be done....


camdendotlol (approved):

jamiefolsom:
@ajolipa — I think this is no longer needed, as v1.9.0 reworked the `tina/config.ts` Clerk setup. A deployed build now throws an error if `TINA_PUBLIC_CLERK_PUBLIC_KEY` is missing, and local dev skips Clerk entirely via `TINA_PUBLIC_IS_LOCAL=true`. If so, should we close this?

PR #623, opened 2026-05-08
Require Clerk SSO for deployed sites; remove Keycloak / username-password auth

ajolipa:
## What

Make Clerk SSO required for every deployed Core Data Places site. Remove the Keycloak and username/password (`tinacms-authjs`) login paths.

## Why

`tinacms-authjs@20.0.1` crashes the TinaCMS admin on every site that doesn't use Clerk — it bundles a React 19 runtime that's incompatible with our React 18 (see #626 for the trace). Fixing it upstream needs a React 19 bump that's blocked elsewhere in our stack.

Rather than carry a broken auth path, v1.9.0 requires Clerk SSO and fails the build fast when Clerk's env vars are missing. Sites that aren't yet on Clerk will fail to build until they're migrated. That's intentional — the failed build is the signal that the site needs migrating.

## What changed

- `tina/config.ts`, `netlify/functions/tina.ts` — require Clerk env vars on deployed builds; throw a clear error if they're missing.
- `tina/auth-provider.ts` — drop the username/password fallback.
- `package.json` — remove `tinacms-authjs` and `@clerk/react`.
- `.env.example` — Clerk vars replace the Keycloak / next-auth ones.
- `docs/upgrade-notes.md` — v1.9.0 entry with the per-site migration steps.

Local development (`TINA_PUBLIC_IS_LOCAL=true`) is unchanged.

## Reviewing

The branch was force-pushed during testing. To get the latest, run `git fetch origin && git reset --hard origin/rb/sso-fix` instead of `git pull`.

You'll see failed deploy-preview checks for any site that isn't on Clerk yet — that's the fail-fast behavior, not a regression. Migration is tracked separately.

Refs #626.


jamiefolsom:
Claude thinks your diagnosis here is right. On develop, tina/config.ts has an unconditional

```ts
const clerk = new Clerk(process.env.TINA_PUBLIC_CLERK_PUBLIC_KEY);
```

The Clerk JS SDK validates the publishable key in the constructor and throws if it's missing/empty, so any site with TINA_PUBLIC_AUTH_USE_SSO unset crashes at module load → tinacms build fails → the site never deploys. 

Lazy-instantiating clerk only when both useSSO and the key are set fixes that — and the defensive clerk? + null-checks in auth-provider.ts are a good complement.

However, the useSSO ? new ClerkAuthProvider(...) logic is commented out, so every non-local site now falls through to UsernamePasswordAuthJSProvider. That breaks Clerk auth for every Clerk-enabled tenant.

Suggested change to tina/config.ts:

```ts
authProvider: isLocal
? new LocalAuthProvider()
: useSSO && clerk
    ? new ClerkAuthProvider({ clerk, orgId: process.env.TINA_PUBLIC_CLERK_ORG_ID })
    : new UsernamePasswordAuthJSProvider(),
```

The useSSO && clerk guard lets a misconfigured SSO deploy (flag set, key missing) degrade to UsernamePassword instead of throwing later at sign-in time.

Let's discuss tomorrow. Let's squash debug commits and add a PR description before merging.

ajolipa:
> Claude thinks your diagnosis here is right.
> Suggested change to tina/config.ts:
> 
> ```ts
> authProvider: isLocal
> ? new LocalAuthProvider()
> : useSSO && clerk
>     ? new ClerkAuthProvider({ clerk, orgId: process.env.TINA_PUBLIC_CLERK_ORG_ID })
>     : new UsernamePasswordAuthJSProvider(),
> ```


Claude is wrong about the diagnosis then, because every iteration that I've tried so far of gating the initialization of the `clerk` instance behind checking for the presence of the environment variables has thrown the same build errors (OOM errors, that is). The code is commented out in the (now second-to-)latest commit because I was (re)verifying that commenting it out causes the builds to complete (it does). So the issue is definitely somehow with the commented out lines, but as you can see from the newest commit, implementing Claude's suggested change brings the build error back. So I remain stumped for the moment, although I can easily believe that the solution will feel obvious once we figure it out.

jamiefolsom (approved):
Tested and good to merge.

Issue #636, opened 2026-05-13
v1.9.0 release tracker


jamiefolsom:
# v1.9.0 release tracker

**Headline:** Clerk SSO becomes required for every deployed CDP site. The Keycloak and username-password (`tinacms-authjs`) paths are removed — `tinacms-authjs` is broken upstream (#626).

## Pre-merge — land on `develop`

- [x] **#628** — Raise Node heap limit for build (prevents `tinacms build` OOM on larger content repos).
- [x] **#623** — Require Clerk SSO; remove Keycloak / username-password.
- [x] **#638** — Honor `GITHUB_BRANCH` in `build.content.mjs` + two defensive null-guards (carries the remaining diff from `deploy/marronage`).
- [x] **#642** — Document and ship the v1.9.0 posts/paths frontmatter migration (`scripts/migrations/v1.9.0-posts-published.mjs` + upgrade-notes subsection covering #635).

## Cut the release

Follow the `release-cdp` workflow.

## Out of scope

- #626 — `tinacms-authjs@20.0.1` React-19 incompatibility (path back is a React 19 bump, separate effort).
- #627 — stray `TINA_LOCAL_CONTENT_PATH` cleanup (handled in the per-site upgrade work).

jamiefolsom:
Released as **v1.9.0**: https://github.com/performant-software/core-data-places/releases/tag/v1.9.0

PR #638, opened 2026-05-13
Honor `GITHUB_BRANCH` when cloning content; add two defensive null guards

jamiefolsom:
## What

Two small, independent improvements:

**Branch-per-environment content cloning.** `scripts/build.content.mjs` now honors `GITHUB_BRANCH` when cloning the content repo. Previously it always cloned the default branch, so a project couldn't keep a separate content branch per deployed environment (for example, `develop` for staging and `main` for production). `.env.example` documents the pattern.

**Two defensive null guards.** Render nothing when a required prop is missing:
- `src/apps/posts/PlaceInsert.tsx` — guards against an unsaved insert in the TinaCMS preview (`place.uuid` not yet set).
- `src/visualizations/Timeline.tsx` — guards against an empty `data` prop.

Both null guards surfaced as runtime crashes during content-editing on marronage staging.

## Testing

`npx vitest run` — 51 passing.

Refs #636.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -12,8 +12,12 @@ CONFIG_URL=https://raw.githubusercontent.com/...
 # If true, Netlify's CDN caching will be disabled
 DISABLE_CACHE=false
 
-# Name of the GitHub branch that will serve as the storage for TinaCMS content
-GITHUB_BRANCH=feature/tina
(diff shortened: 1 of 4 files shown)


ajolipa (approved):
Code looks good; assuming you've tested the `build.content` script and it all seems good, then go ahead and merge.

PR #648, opened 2026-05-20
Contain post-embed crashes so removing a map doesn't kill the editor

### LINKED ISSUE #647: Bug: Visual editor crashes when removing a map visualization
Removing a map embed crashes the whole page and requires it to be reloaded (without changes having been saved).

https://github.com/user-attachments/assets/d3f6a3de-99c8-4886-aee2-bcef88b40158

(Unclear if this is a result of the updates in PR #646 or whether this bug was never effectively squashed, but just wanted to record that it's still happening.)

jamiefolsom:
## Summary

Closes #647. Removing a `<map>` embed from a post in the TinaCMS visual editor crashed the whole editor — both the preview and the form went blank, a reload was required, and unsaved edits were lost. This wraps the post body in an error boundary so a single embed's crash is contained, the editor stays usable, and the preview recovers on its own.

## Root cause

The post body renders every embed (`<map>`, `<place>`, `<timeline>`, …) through `TinaMarkdown` in `PostContent.tsx`, mounted as a `client:only` island with no error handling. Removing a map unmounts a MapLibre map whose layer/image children clean up against an already-torn-down style:

```
TypeError: Cannot read properties of undefined (reading 'removeImage')
    at Map.removeImage (maplibre)
    at safelyCallDestroy → commitHookEffectListUnmount
    at commitPassiveUnmountInsideDeletedTreeOnFiber → … → flushPassiveEffects
```

It's a throw in a `useEffect` cleanup during unmount, in `@peripleo/maplibre`. With no error boundary, React unmounts the entire root, which also tears down the visual-editing bridge and clears the form. Our visual editor is probably an outlier in terms peripleo integrations, but I'll discuss with @aboutgeo whether a fix could make its way upstream.

This is **pre-existing**, not introduced by #646: it reproduces both with and without that PR's per-instance `MapProvider`. (Without it, the same teardown throws `map.getStyle().layers` instead — the cross-contamination crash the `MapProvider` change was added to fix.)

## The fix

`PostEmbedErrorBoundary` wraps `<TinaMarkdown>`. React routes commit-phase errors (including effect-cleanup throws) to the nearest *mounted* ancestor boundary, so the editor island stays mounted — no blank page, no reload, **edits are preserved**. The boundary wraps the whole body deliberately: a per-embed boundary inside the removed block is itself being deleted and can't catch its own teardown throw — only an ancestor can.

It also recovers on its own:

- **Auto-retry.** On catch, the boundary schedules a single deferred, de-duplicated retry. Removing a map throws several times in one synchronous burst (the removed map plus index-keyed siblings that remount), so the retry waits for the burst to settle and runs once, re-rendering the corrected content. The retry budget counts *bursts* and is capped, so a genuine render-phase error (throws every render) stops on the fallback after a few attempts instead of looping into a root unmount. The retry only re-renders from current content and never touches form state, so edits are never lost.
- **resetKeys.** Keyed on the post body, so the budget replenishes and the boundary resets on each content edit even if auto-retry was exhausted.
- **Context-aware fallback copy.** The fallback is, in practice, only hit inside the visual editor, so it shows a reassuring "your edits are safe" message there and a neutral one on a normally-loaded published page.

## Known behavior / follow-up

The boundary *contains and recovers from* the throw; it doesn't prevent it. Eliminating the throw at the source needs an upstream guard on the `@peripleo/maplibre` layer/image cleanup (e.g. `if (map?.style)` before `removeImage` / `getStyle().layers`) — worth filing separately; out of scope here to keep this tightly scoped.

## Verification

- Reproduced the crash and verified the fix in the live visual editor (Playwright) on a post with multiple map embeds: before — both panels blank, reload needed; after — editor stays mounted, form intact with the removal applied, and the preview auto-recovers (the fallback flashes briefly during the teardown burst, then the maps re-render).
- `npm run vitest` green (122 passed, 3 skipped), including `test/postEmbedErrorBoundary.test.ts` covering the boundary's error/reset logic.

## Test plan

- [ ] Open a post with a `<map>` embed in the visual editor; remove the map. Editor stays mounted; no reload; the preview recovers; the change can be saved.
- [ ] Remove the first of several embeds (forces trailing remounts); editor survives and recovers.
- [ ] Live published post with maps renders normally; fallback does not appear in the happy path.

### DIFF
--- a/src/apps/posts/PostContent.tsx
+++ b/src/apps/posts/PostContent.tsx
@@ -1,6 +1,7 @@
 import PlaceInsert from '@apps/posts/PlaceInsert';
 import IframeEmbed from '@components/IframeEmbed';
 import MediaInsert from '@components/MediaInsert'
+import PostEmbedErrorBoundary from '@components/PostEmbedErrorBoundary';
 import TranslationContext from '@contexts/TranslationContext';
 import { useTranslations } from '@i18n/useTranslations';
 import { Peripleo as PeripleoUtils } from '@performant-software/core-data';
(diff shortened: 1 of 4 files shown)


ajolipa (approved):
This looks great!

PR #677, opened 2026-06-16
Full view for paths

camdendotlol:
# Summary

This PR:

* adds a new `view` setting for paths.
  * `zoom` (default) displays paths the same way as before
  * `full` renders all places on the map within the same view with a hardcoded buffer and does not pan/zoom the map as you navigate between places
* adds blue styling to indicate the currently selected place in full mode
* factors out some code from the `PlacesMap` component into the `usePlacesFeatures` hook so we can reuse the functionality in `PathViewer`
* wraps a bunch of `svg` and `div` elements that had `onClick` props into Headless UI `Button`s to improve accessibility without any visible changes
* fixes an issue where the back button was not always properly grayed out on the path viewer's starting screen

### Points-based map with a place selected
Screenshot 2026-06-16 at 3 15 19 PM

### Polygon map
Screenshot 2026-06-16 at 3 02 18 PM


blms (approved):
Looks good! Good ideas with the new hook and accessible buttons.

Is the deno lockfile intentional?

(By the way, I wonder if we should add `config.json` to `.gitignore` at this point, considering we have fairly complete documentation about what needs to go in it, and it's annoying to have to deal with its automated changes in git)

camdendotlol:
> Is the deno lockfile intentional?

Nope - it showed up after running `netlify dev` for the first time on this branch. Netlify runs Deno under the hood for edge functions, so I assumed it was a new thing in the latest Netlify CLI release.

camdendotlol:
> (By the way, I wonder if we should add config.json to .gitignore at this point, considering we have fairly complete documentation about what needs to go in it, and it's annoying to have to deal with its automated changes in git)

That sounds good to me. Maybe we could rename the existing on to `config.example.json` and leave it checked in.

blms:
> That sounds good to me. Maybe we could rename the existing on to `config.example.json` and leave it checked in.

Sure!

PR #687, opened 2026-06-17
Add overlay option to Tina for paths

camdendotlol:
# Summary

This PR adds a new option to the Tina form for paths where the user can select a layer from the `layers` array in `config.json`.

## Screenshot
Screenshot 2026-06-17 at 4 27 24 PM


blms (approved):
Looks good! Great to have support for these in paths.

PR #696, opened 2026-06-22
Show all geometry on zoom paths

camdendotlol:
# Summary

This PR updates zoom paths (i.e. the default, original paths before [full paths were added](https://github.com/performant-software/core-data-places/pull/677) to continue rendering all geometry and lines when a place is selected. The only difference between full and zoom mode now is that zoom mode automatically centers/zooms the map on the selected place.

Also, this PR cleans up map-related conditionals in `PathSelectionManager` that were breaking the selection state for zoom mode and seemed to be unnecessary anyway.


blms (approved):
Looks good!

Issue #655, opened 2026-06-04
Bug: map does not zoom to a selected location


jamiefolsom:
Selecting a specific record (e.g. clicking a search result) does not fly/zoom the map to that location. Reported by the Marronage project (June 4, 2026); flagged as a priority fix.

This may share a root cause with #145 (map not zooming to data points), but the trigger differs: #145 is an initial-load fit-to-data timing issue, whereas this is **selection-driven** zoom. Cross-linking for tracking.

Related: #145, #255, #256.

From: Marronage client meeting, June 4, 2026.

PR #700, opened 2026-06-23
Fix map search never fitting bounds to results or selection

jamiefolsom:
## Problem

On the map-based search, result points render but the map **never moves the camera**: it doesn't zoom to the result set on load, and clicking a result doesn't pan or visibly highlight it. The map stays at world view, so selections look like no-ops.

## Root cause

`src/components/Map.tsx` wraps `<PeripleoMap>` in its own nested `<MapProvider>` (introduced to isolate multiple maps in a post body so they don't overwrite each other's `setMap`).

But `MapView` and `MapSearchContext` render **above** `<Map>` and read the map via `useLoadedMap()` against the **root** `MapProvider` that `<Peripleo>` supplies. With the nested provider in place, `<PeripleoMap>` registers the map into the *inner* context, so those ancestors receive `null` permanently. Consequently:

- `MapSearchContext.getBoundingBox()` hits `if (!map) return;` and returns a Promise that never resolves, so `.then(map.fitBounds)` never runs.
- `MapView`'s fit-bounds effect guard (`… && map && …`) is never satisfied.

The result layers still render because they are **children** of `<Map>` and see the inner context — which is why the bug is silent (dots show, camera never fits).

Verified on the deployed staging build: `getBoundingBox()` times out, and the map makes zero camera calls (`fitBounds`/`easeTo`/`flyTo`/`jumpTo`).

## Fix

Add an `isolate` prop to `<Map>` (default `true`, preserving the multiple-maps-in-a-post isolation) and set `isolate={false}` on the search map so it registers into the root `MapProvider` that its ancestors read.

## Scope

- Default `isolate={true}` means detail-page maps, paths, and post embeds are unchanged.
- Only the search map (`MapView`) opts out; the search renders a single map, so it doesn't need per-map isolation.

## Verification

- `tsc` passes for the changed files.
- Root cause confirmed at runtime on staging (getBoundingBox hang, zero camera calls, `useLoadedMap()` null in the ancestors).
- Recommend confirming on the deploy preview: the place search should zoom to the result cluster on load, and clicking a result should pan/highlight it.

### DIFF
--- a/src/apps/search/map/MapView.tsx
+++ b/src/apps/search/map/MapView.tsx
@@ -57,6 +57,7 @@ const MapView = () => {
       classNames={{
         controls: controlsClass
       }}
+      isolate={false}
     >
       { layerType === LayerTypes.single && (
         <SingleLayer
--- a/src/components/Map.tsx
+++ b/src/components/Map.tsx
@@ -48,6 +48,20 @@ interface Props {
     controls?: string
     root?: string,
   };
+  /**
+   * When `true` (the default), the map is wrapped in its own `MapProvider` so it
+   * gets an isolated `MapContext`. This is required when several maps are mounted
+   * under the same Peripleo root (e.g. a post body with multiple map blocks),
+   * where a shared context would let each `PeripleoMap` overwrite the previous
+   * one's `setMap(...)`.
+   *
+   * Set to `false` when the consumer renders a single map and has ancestor
+   * components (outside `<Map>`) that need to read the map via `useLoadedMap()` —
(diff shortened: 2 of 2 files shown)


camdendotlol (approved):
I've tested this locally and it looks like a reasonable short-term fix for https://github.com/performant-software/core-data-places/issues/655 that doesn't seem to break anything in the places where `Map` is used.

It does feel a bit hacky to have a conditionally-rendered context wrapper so we might want to refactor a bit more down the road.

Issue #708, opened 2026-06-24
Clean up package.json dependencies


jamiefolsom:
Removing `netlify-cli` (#707) cut a large chunk of the dependency tree. A scan for similar cases turned up a few more `package.json` cleanups. This issue tracks the rest; #707 covers `netlify-cli` itself.

Verified two ways: `depcheck` plus a manual scan that also reads `.astro` and CSS files (depcheck misses both). Only items both methods agree on are listed as removable.

### Unused — safe to remove

Not imported anywhere in code, not used by any script:

- `@turf/bbox` — only `@turf/helpers` and `@turf/truncate` are imported
- `@netlify/functions` — the Tina function is built on `serverless-http` + `express` and never imports it
- `typesense-instantsearch-adapter` — never imported; Typesense helpers come from `@performant-software/core-data`

Confirm with a build and a deploy before removing, since a wrong call here would break the CMS function quietly.

### Imported but not declared

These are imported in code but missing from `package.json`. They only resolve because another package pulls them in, so they break if that package changes. Add them as direct dependencies:

- `maplibre-gl` — `src/utils/search.ts`, `src/utils/map.ts`
- `@turf/turf` — `src/apps/paths/PathViewer.tsx`
- `@tinacms/schema-tools` — `tina/content/*.ts`

### Housekeeping

- `@tailwindcss/typography` is listed in both `dependencies` and `devDependencies` — drop the `devDependencies` copy
- Build/type-only tools sit in `dependencies` and could move to `devDependencies`: `typescript`, `@types/react`, `@types/react-dom`, `@astrojs/ts-plugin`, `@tinacms/cli`

PR #721, opened 2026-06-25
Dependency upgrades and performance improvements

camdendotlol:
# Summary

This PR upgrades a bunch of dependencies, most notably Astro, and includes a couple minor performance fixes.

## Dependencies
- Vite 8
- Astro 7 (including some tweaks to make our content collections work with their new content layer system, mostly related to stricter checks rather than actual API changes)
- TypeScript 6
- updated EditionCrafter to the release version of 1.3.1 instead of the beta
- bumped `underscore` and `swiper` due to CVEs that WebStorm warned me about
- removed `@turf/bbox` from `package.json` (which doesn't really do anything because `@turf/turf` includes all Turf packages anyway)
- removed `@netlify/functions` and `typesense-instantsearch-adapter` from `package.json` as per https://github.com/performant-software/core-data-places/issues/708
- pulled the duplicate `@tailwindcss/typography` out of dependencies leaving the one in `devDependencies`
- updated Nanostores to the latest version
- removed `dotenv` from `package.json` and replaced it with Node's native `loadEnvFile`
- removed `uuid` and replaced it with the browser's built-in `randomUUID`

## Performance fixes
- moves the Swiper CSS import from `Layout` to the actual Swiper component so we only send it to the user when it's needed
- adds a cache to the `i18n` backend helper function. In theory `useTranslations` only fetches this once, but there seem to be multiple places where separate translation contexts live in multiple parts of the component tree and I confirmed via the console that duplicate fetch requests are hitting the cache now

## Other
- adds missing key props in the Mobile Header that were displaying errors on every page


blms (approved):
Looks good!

PR #723, opened 2026-06-29
Tina upgrades and caching

camdendotlol:
# Summary

This PR implements a simple caching layer in `backend/tina`. The cache only exists per-page since we're serving the site with Netlify functions, but there are places where we fetch content such as translations and the navbar in multiple SSR components on the same page.

This PR also upgrades all Tina-related dependencies to their latest versions and removes a couple unused functions from `backend/tina`.


blms (approved):
Looks good!

Issue #211, opened 2025-05-19
Bug: Tina Media Insert not loading for some records


ajolipa:
When a manifest is generated or re-generated in Core Data, in some cases the `id` field of the manifest does not actually match the URL of the manifest, which causes an error in newer versions of the Clover IIIF viewer (2.11+?) that leads the component not to load. (See https://github.com/performant-software/iiif-cloud/issues/94)

PR #725, opened 2026-06-29
More dependency upgrades

camdendotlol:
# Summary

This PR upgrades almost all the dependencies that haven't been upgraded yet. The ones that required some level of code changes or explanation are:
- Clerk packages. They were very outdated and the new ones had a different API.
- Express 4->5 had some fairly minor changes to how it handles wildcard routes and request attribute modification
- Replaced our own (I'm pretty sure copy/pasted from somewhere else) `db-provider` with Tina's own `mongodb-level`, which also involved removing the `abstract-level` package
- Replaced a couple Turf-related packages with `@turf/turf`, which includes them all and should make upgrades simpler.
- Removed `vite` from `package.json`. Astro uses Vite internally and does not expect us to include it in our top-level dependency list. The only reason we had it there was because we import `loadEnv` from it in the Astro config. That import still works since Astro pulls in Vite. It's not always ideal to import from a package that's not in your `package.json`, but having `vite` there might lead to weird discrepancies/conflicts between the one we're pulling in and the one Astro is actually trying to use. Vite upgrades tend to be major features of new Astro releases.

## Skipped packages

This will probably be my last cleanup/upgrade PR unless I find something else, and there are a few packages that I did not upgrade:
- React is impossible to bring to version 19 without a lot of work. The error I see right now is a conflict with EditionCrafter, but I'm pretty sure `react-components` also conflicts.
- Clover continues to be blocked from upgrading by https://github.com/performant-software/core-data-places/issues/211

## Testing notes

There are a few areas of concern once this is deployed to staging:
- [x] Tina rebuild button
- [x] Tina image uploads
- [x] make sure list/card search still works
- [x] make sure the image carousel still works on both desktop and mobile (there's one on PADP)


blms (approved):
Looks good!

PR #727, opened 2026-06-30
Tina/Clerk fixes

camdendotlol:
# Summary

This PR fixes a couple issues:
- a regression from #725 where the updated Clerk `authenticateRequest` calls weren't quite right (the new API also requires the publishable key to be passed to it)
- a pre-existing issue with the rebuild button where it only checked permissions against the user's active org, which may not be the same as the org that Tina is connected to. To fix this, I updated the Tina auth handling to set the user's active org to match the `TINA_PUBLIC_CLERK_ORG_ID` when they log into Tina.


blms (approved):
Looks good!

PR #732, opened 2026-07-10
Fix inverted condition in getDetailPagePaths

### LINKED ISSUE #729: Static builds create detail pages for the wrong models
In a static build (`STATIC_BUILD=true`), detail pages are built for exactly the models that are **not** listed in `config.detail_pages.models`, and skipped for the models that are.

The cause is an inverted condition in `getDetailPagePaths`, in `src/utils/detailPages.ts`:

```js
if (hasDetailPage(model)) {
  return routes;   // routes is still the empty list here
}
```

A model with a configured detail page returns zero routes. Every other model falls through and generates a route per record.

Sites in production do not show this, because they run the server output, which never calls `getStaticPaths`. Detail pages render on request there, so only static output is affected.

Observed on v1.10.0 with a site whose configuration lists `people`, `places`, and `works`. The static build produced `en/events/` and `en/organizations/` pages instead. With the condition flipped, the configured models build correctly: every place, person, and work page is present, and none appear for the unconfigured models.

Fix:

```js
if (!hasDetailPage(model)) {
  return routes;
}
```

jamiefolsom:
`getDetailPagePaths` returned an empty route list for models that **have** configured detail pages, so a static build (`STATIC_BUILD=true`) prerendered detail pages for exactly the wrong set of models. SSR is unaffected because `getStaticPaths` is never called there.

One-line fix: negate the `hasDetailPage` guard. Verified on a static build of an internal site — the configured models (people, places, works) now get detail routes and the unconfigured ones don't.

Fixes #729

### DIFF
--- a/src/utils/detailPages.ts
+++ b/src/utils/detailPages.ts
@@ -7,7 +7,7 @@ import { hasDetailPage } from '@utils/detailPagePaths';
 export const getDetailPagePaths = async (model: Models) => {
   let routes = [];
 
-  if (hasDetailPage(model)) {
+  if (!hasDetailPage(model)) {
     return routes;
   }
 


blms (approved):
Good catch!

PR #746, opened 2026-07-13
Configurable CDN cache TTLs + purge-on-publish

### LINKED ISSUE #744: Pages are CDN-cached for only 5 minutes, and publishing content doesn't invalidate the cache
All SSR cache policy is one hardcoded header in `setCacheControl` (src/utils/url.ts): 5 minutes fresh, then stale-while-revalidate. Content publishes don't purge anything — a Tina save is only visible when the TTL runs out.

The result is the worst of both: low-traffic sites age out of cache between visits, so most real visitors wait for full server renders (measured 0.5–7 s per server island, and 6–10 s whole-page SSR on a cold site), while editors still wait up to 5 minutes to see their changes.

Proposed fix (PR to follow):
- Make the CDN TTL and stale-while-revalidate values configurable per site via env vars, defaulting to today's values so nothing changes without opt-in.
- Add a `Netlify-Cache-Tag` to cached responses and call `purgeCache` (already available in our `@netlify/functions` version) when TinaCMS saves or deletes content, so a publish is visible in seconds regardless of TTL.

With purge-on-publish in place, a site can safely run a long TTL and serve nearly every visitor from the edge (~25 ms) instead of re-rendering.

jamiefolsom:
Closes #744.

Two changes, both inert until a site opts in:

1. **Configurable TTLs.** `CACHE_CDN_MAX_AGE` and `CACHE_STALE_WHILE_REVALIDATE` env vars feed the `Netlify-CDN-Cache-Control` header, defaulting to the current hardcoded values (300 / 604800) — with no env vars set, the emitted headers are byte-identical to today's. Browser-facing `Cache-Control` is unchanged.

2. **Purge-on-publish.** Cached responses now carry a `Netlify-Cache-Tag`, and TinaCMS saves/deletes call `purgeCache` for that tag after the GitHub write (`tina/git-provider.ts`), so a publish is visible in seconds regardless of TTL. A failed purge logs and never fails the save; local dev never purges (`GitHubProvider` only exists when not local). This is what makes a long TTL safe: the plan is to trial `CACHE_CDN_MAX_AGE=86400` on one production site and measure against the baselines taken today.

Record-detail islands set no cache headers, so live FairData content is unaffected by any TTL.

**Tests:** new `test/cacheControl.test.ts` (defaults byte-identical, env overrides, tag present, `DISABLE_CACHE` short-circuit); full vitest suite green (69 passed).

**One thing to validate on staging before any site opts in:** whether `purgeCache` picks up an ambient token in our v1-style function; if it 401s, the site needs `NETLIFY_PURGE_API_TOKEN` set (will document either way).

### DIFF
--- a/astro.config.mjs
+++ b/astro.config.mjs
@@ -48,6 +48,18 @@ export default defineConfig({
   },
   env: {
     schema: {
+      CACHE_CDN_MAX_AGE: envField.number({
+        access: 'public',
+        context: 'client',
+        default: 300,
+        optional: true
+      }),
+      CACHE_STALE_WHILE_REVALIDATE: envField.number({
+        access: 'public',
+        context: 'client',
+        default: 604800,
+        optional: true
+      }),
       DISABLE_CACHE: envField.boolean({
         access: 'public',
         context: 'client',
--- a/src/env.d.ts
+++ b/src/env.d.ts
@@ -9,6 +9,8 @@ declare module '@performant-software/shared-components';
 declare module 'underscore';
 
 declare module "astro:env/client" {
+  export const CACHE_CDN_MAX_AGE: number;
+  export const CACHE_STALE_WHILE_REVALIDATE: number;
   export const DISABLE_CACHE: boolean;
   export const PRELOAD_MAP: boolean;
   export const STATIC_BUILD: boolean;
--- a/src/utils/url.ts
+++ b/src/utils/url.ts
@@ -1,4 +1,4 @@
-import { DISABLE_CACHE } from 'astro:env/client';
+import { CACHE_CDN_MAX_AGE, CACHE_STALE_WHILE_REVALIDATE, DISABLE_CACHE } from 'astro:env/client';
 import _ from 'underscore';
 
 interface SearchParams {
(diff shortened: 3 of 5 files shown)


jamiefolsom:
We tested this on the padp-staging deploy preview, against the published develop deploy as a baseline. Staging previously had DISABLE_CACHE=true, so caching was off entirely; we removed that first so the baseline matches production's normal five-minute caching.

While the cache is fresh, the two behave the same: repeat requests come back from the CDN in under 100 ms. The difference shows up after five minutes. The baseline's cached copy expired on schedule and requests went back to the server. The preview, built with CACHE_CDN_MAX_AGE=86400, was still served from the CDN a few minutes later. The longer cache lifetime works.

One part can only be tested after this merges: clearing the cache when an editor publishes. The cache-clearing API only accepts a token that Netlify provides to deployed functions, not personal tokens, so I couldn't call it from outside. Once staging's published deploy includes this change, saving a page in TinaCMS will show whether the cache clears right away, and whether our function receives that token automatically or the site needs it set by hand.

ajolipa (approved):

Issue #743, opened 2026-07-13
Server islands return a 500 error when the request has no Referer header


jamiefolsom:
A server island is a part of a page that the browser fetches from the server after the page itself has loaded. On v1.10.0 and on older deploys, any `/_server-islands/*` request that arrives with no Referer header returns a 500 error. The same URL with a same-origin Referer returns 200 and the full content. Browsers always send a Referer on these fetches, so visitors see working pages. Command-line requests, uptime monitors and scripted checks see 500s, and that is how this was first misread as a user-facing outage.

This replaces the original diagnosis, which blamed a mismatched React version.

To reproduce on any deployed site:
```
# extract an island URL from any page's HTML, then:
curl -s -o /dev/null -w '%{http_code}\n' "$SITE$ISLAND_URL"                       # 500
curl -s -o /dev/null -w '%{http_code}\n' -H "referer: $SITE/en/" "$SITE$ISLAND_URL"  # 200
```

Likely cause: the island reads the locale and route from the Referer through Astro's internationalization support, and throws when the header is missing instead of falling back to the default locale. The server logs show a React "Invalid hook call" warning on the failing path, but that warning comes from the error path rather than from the cause.

Two suggested fixes:

- Fall back to `i18n.default_locale` when there is no Referer and no locale can be worked out, instead of throwing.
- In `Post.astro`, `if (publishedOnly && !post.published)` still raises a TypeError when `post` is null, which is what happens for a missing record. A null check would make that a 404 instead of a 500.

One more source of log noise: a site with no custom navbar logs `NotFoundError: Unable to find record content/navbar/en.json` on every header render. The default navbar renders correctly, but the message is recorded as an error.

PR #747, opened 2026-07-13
Render Header and Footer statically instead of as server islands

### LINKED ISSUE #745: Header and Footer are server islands, but their content only changes on publish
Every page view fires separate server-island requests for the Header and Footer. Both render navigation, translations, and branding that change only when someone publishes in the CMS — there's nothing per-request about them except the active-nav highlight, which is derived from the request's Referer header because an island doesn't know what page it's on.

Costs of the current shape:
- Two extra SSR round-trips per page view; on cache misses each takes 0.15–2.2 s (measured), so navigation pops in late on cold visits.
- The Referer dependency means island requests without that header fail (#743), and the header/footer never resolve at all in static builds.

Proposed fix (PR to follow): render Header and Footer directly in the page HTML and pass the page's own pathname down for active-nav state. The page response is already CDN-cached per URL, so each cached page carries the right active state; the transparency wrapper (`HeaderContainer client:load`) is unaffected. Content islands (posts, detail pages, home) stay as they are.

jamiefolsom:
Closes #745. **Draft — intended to land after #746** (the cache TTL/purge PR), whose CDN caching absorbs the cost of moving the navbar queries into the page render.

Header and Footer content (nav, translations, branding) only changes on CMS publish, but as server islands they cost two extra SSR round-trips per page view — 0.15–2.2 s each on cache misses in this week's measurements — and they fail entirely when an island request has no Referer header (#743), since that's where an island learns what page it's on.

Changes:
- `Layout.astro`: `server:defer` removed from Header and Footer; each receives `pathname={Astro.url.pathname}`.
- `Header.astro` / `Footer.astro`: active-nav state and the language picker use the `pathname` prop instead of the Referer; their per-island `setCacheControl` calls are gone (the page response, which now contains them, is already covered by Layout).
- `getCurrentURL` deleted from `src/utils/url.ts` — these were its only callers.

Not touched: `HeaderContainer client:load` (transparency wrapper), content islands (posts, paths, pages, detail pages, home).

Why prop threading instead of client-side `location.pathname`: NavBar renders on the server, and the CDN cache key is the URL, so every cached page carries its own correct active state — including under `STATIC_BUILD`, where header/footer previously never resolved at all (islands need a server).

**Verified:** vitest suite green; components compile in a static build (which then fails at a pre-existing, unrelated `getStaticPaths` crash in `api/gallery/[manifestId].json.ts` when the config has no gallery — identical failure on develop; separate issue to follow). Render-level QA: deploy preview — check a custom-navbar site and a default-navbar site, active-nav highlighting, mobile header, and the language picker on a multi-locale site.


ajolipa (approved):
This looks great to me.

ajolipa:
I suppose before merging, in addition to resolving conflicts, I'd like to understand why some of the deploy previews (but not others) are just rendering blank pages (see the `juel-staging` and `juel-ancestry` previews, possibly among others). It might be unrelated to this PR specifically but I just want to know for sure.

PR #751, opened 2026-07-14
Page Builder Tina Visual Editing

ajolipa:
### In this PR
Add visual editor functionality to the Tina page builder! The major steps involved in this are as follows:
- Add the `@tinacms/astro` integration;
- Create an island registry (currently only has the entry for `page`) to tell Tina how to re-render the body of a page while editing it. This involves updating the API calls to include the `fetchWithMetadata` wrapper so that the Tina code has what it needs to map between the form and the page components;
- Annoyingly, the whole thing breaks if there are any TSX components inside the island that Tina is managing the rerendering of, so I did a bit of refactoring and passing around an `isTinaPreview` flag to handle rendering a trimmed-down version of the header in the visual editor; this enabled the preview to update accurately when the "transparent navbar" field is toggled in the editor, which was otherwise causing the refresh to fail.
- The `isTinaPreview` flag is also used in the `PageBody.astro` component in a few cases to determine whether to render a React component or a static preview image;
- The `TextImage` component is reworked as an Astro component with some vanilla JS controlling the image resizing; this allows us to avoid maintaining a bunch of duplicate components in both Astro and TSX form;
- There are a couple of components that I couldn't avoid having multiple versions of, namely `Tabs` and `Card` -- `Card` because it is used both in the page builder and on the post/path index pages, and `Tabs` because a generic preview image did not seem sufficient for the visual editor in this case, so instead I made an Astro version that just stacks the tab content previews on top of each other. In both of these cases, I grouped the components into a subfolder and added a `styles.ts` file that contains all of the actual Tailwind classes applied within the component. Definitely open to better ideas about how to handle this part, but at least this avoids just having duplicate CSS that has to be edited twice in any future updates.

Demo video:

https://github.com/user-attachments/assets/2dc01d35-48ee-4e42-87aa-6167b9095e54


camdendotlol (approved):
The code looks good to me but I did not run it so I suggest waiting for Ben's approval too.

src/assets/pages/image_carousel.png
blms, line 1: Could we use something like [tinypng](https://tinypng.com/) to compress these and/or could they be lower quality? 1.29MB seems like a big file to load on each preview with an image carousel

src/apps/pages/TextImage.astro
```diff
@@ -0,0 +1,164 @@
...
+      />
+    </div>
+  </div>
+</div>
+
+<script lang="ts">
+  const IMAGE_SCALE = 1.75;
+  const BACKGROUND_SCALE = 1.65;
+  const SCREEN_WIDTH_LG = 1024;
+  const instanceState = new WeakMap();
```
blms, line 109: Nice, I had never heard of `WeakMap` before!

src/components/Column.astro
```diff
@@ -31,7 +32,7 @@ const { items, className, url, t } = Astro.props;
     )}
   >
     { _.map(items, (item) => ( 
-      <ColumnContent item={item} />
+      <ColumnContent item={item} isTinaPrevew={isTinaPreview} t={t} />
```
blms: `isTinaPrevew` typo?

blms (approved):
Code looks great! I like just showing the images for the components that can't be rendered in preview—just a minor question about asset size for those. I also think `styles.ts` is a great solution to the shared CSS issue 👍  

Also tested it out locally with a few different pages and block types and this is looking great, like a full-featured WYSIWYG page builder at this point. Really nice work @ajolipa.

src/apps/pages/TextImage.astro
ajolipa: Haha I admit I was pulling my hair out for a while trying to debug the behavior when there's more than one copy of the component. ^_^;; There might be a simpler way to keep track of stuff but at least this seems to work!

PR #760, opened 2026-07-21
Make Tina content static

### LINKED ISSUE #733: published filter is ignored when building post and path routes
`getStaticPaths` in `src/pages/[lang]/posts/[slug].astro` and `paths/[slug].astro` passes `published: { eq: true }` at the top level of the Tina query params, where it is silently ignored — the filter belongs inside `filter:`.

The result: static builds create public routes (and sitemap entries) for **unpublished** posts and paths. The page body then 404s out of the `publishedOnly` check in the Post/Path component, so each unpublished item ships as a live URL with an empty body. On one internal site, all 12 posts are unpublished and all 12 got public routes.

Fix: wrap the condition in `filter:` in both files, e.g.

```js
const { posts } = await fetchPosts({
  first: -1,
  filter: { published: { eq: true } }
});
```

SSR is unaffected (`getStaticPaths` is not used there).

### LINKED ISSUE #758: Separate saving content from publishing it
As an editor, I want to save and preview changes without them appearing on the public site, and publish them deliberately when I'm ready.

Today a Tina save is public immediately: the save commits to the content repo, updates the datalayer, and the next request renders it. Now that the visual editor (#751) covers preview before and after save, immediate publication is no longer needed for editors to see their work.

Proposed direction: prerender the Tina content routes (pages, posts, paths) at build time. The preview route added in #751 stays live for editors; the existing "Rebuild site" button becomes the publish action.

Related: #733 — route building must respect published state for this to be correct.

### LINKED ISSUE #759: Console error on pages with TextImage: 'IMAGE_SCALE' has already been declared
Pages that use the TextImage page-builder component now log an error in the browser console:

`Uncaught SyntaxError: Failed to execute 'before' on 'Element': Identifier 'IMAGE_SCALE' has already been declared`

We see it on the home pages of padp-staging and marronage-staging, which run the latest develop. PADP production runs v1.10, which predates #751, and does not show the error.

What seems to happen: #751 rewrote TextImage as an Astro component with a small script that resizes the image. The script declares a constant named `IMAGE_SCALE`. It runs once when the page loads, and runs again when the server island's HTML is inserted into the page. The second run finds the constant already defined and throws.

The pages still look fine, but the resize script probably never runs for the images inside the server island.

Two possible fixes: either wrap the script in a function so the constant is not global, or have the script check whether it already ran.

ajolipa:
### In this PR
- Enables the `i18n` content collection by default, meaning that localization data will be fetched at build time as a snapshot and fetched using Astro's content collection API rather than live calls to the Tina backend;
- Add a `navbar` content collection and enable it by default, so that the navbar content on each page is being fetched via the Astro content API rather than fetched from Tina on every page;
- Eliminates the `HeaderContainer` component, since the `transparent` prop can now be passed directly to `Layout` on all pages;
- Adds `prerender = true` to all `pages`, `posts`, and `paths` content, and removes the server island directives (which have no meaning on SSG pages);
- Miscellaneous other small tweaks with the intention of smoothing out page loads;
- A couple of bug fixes that surfaced during this work (see Issue #733 and Issue #759 .

The idea behind these updates is that deployed sites will fetch data from Tina only during build, and never at the time of page load; this ought to drastically help performance. To preview paths, posts, and pages, users can make use of the visual editor in Tina. When they're happy with changes they can save their work and then redeploy the site using the button on the Settings collection page. This offers a more controlled editing experience, rather than having all Tina content updates go live as soon as they're saved.

Closes #758 . Closes #759 . Closes #733 .


jamiefolsom:
This implements #758: content routes prerender at build time, the visual editor's preview route stays live for editors, and the rebuild button becomes the publish step. It also fixes #733 — the published filter was passed outside `filter:` and silently ignored, so unpublished posts and paths were getting public routes; this PR nests it correctly. Worth adding `Closes #758, closes #733` to the description so both close on merge.

camdendotlol (approved):

PR #762, opened 2026-07-22
Fixing FOUC layout issues

### LINKED ISSUE #761: Bug: Bundled CSS not injected into the `<head>`
Because of the presence of a `client:load` component inside the `<head>` in the `Layout.astro` file (namely, `<Title>`), Astro fails to add the generated `<link rel="stylesheet"... />` bundled styles inside the `head`, and they end up in the `body` instead. The reason has to do with the client directive leading to an `astro-island` element being placed in the HTML `head`, which is not an allowed element; because the `head` is then not syntactically well formed, the styles get injected in the wrong place. Because the styles are added in the `body`, the browser will try to render the page header before the stylesheets have loaded, causing a flash of unstyled content.

ajolipa:
### In this PR
Resolves #761 by removing the client-loaded `<Title>` component from the `Layout`, and instead either passing the page title as a direct prop when possible, or using vanilla JS to update the title from within a child component when necessary. This allowed for some cleanup of some unnecessary components and stores.


blms (approved):
Looks good!

PR #765, opened 2026-07-24
Mobile search layouts

ajolipa:
### In this PR
Improvements to the mobile layout of the map and list searches.

image

image

image

image

image

image

### Notes
As a first pass, I just made it so that the table and timeline views aren't available on small screens, since developing versions of those that look good on small screens seems like a bigger list. Open to suggestions though.


src/apps/search/list/ListLayout.tsx
```diff
@@ -3,49 +3,69 @@ import { SearchBox } from 'react-instantsearch';
...
 
   const { t } = useContext(TranslationContext);
 
+  useEffect(() => {
+    if (window.innerWidth < MEDIUM_SCREEN_SIZE) {
+      setFilters(false);
+    } else {
+      setFilters(true);
+    }
+  }, [window.innerWidth]);
```
blms: I was wondering about this because I haven't seen `window.innerWidth` used in a `useEffect` before. Does this have the same general behavior as a resize observer/resize event listener? I wonder if it might be more stable to wrap the logic in an event handler and then call that on the `resize` event, attaching the event listener on mount and removing it on unmount. In other words something like:

```suggestion
  useEffect(() => {
    const handleResize = () => {
      if (window.innerWidth < MEDIUM_SCREEN_SIZE) {
        setFilters(false);
      } else {
        setFilters(true);
      }
    }
    window.addEventListener('resize', handleResize);
    return () => window.removeEventListener('resize', handleResize);
  }, []);
```

blms (approved):
Looks great! Just one question about resize in `useEffect`.

src/apps/search/list/ListLayout.tsx
ajolipa: Oh yeah that is definitely better. 😂 For some reason I momentarily blanked on the existence of resize observers. Will fix!

PR #766, opened 2026-07-26
Add `Edit History` field on Tina collections

ajolipa:
### In this PR
Adds a common field to every Tina collection that uses a `beforeSave` hook to keep track of a basic edit history of every document.

*Logged in as Admin*
(Defaults to showing the most recent five entries; expands with the button to show all.)
image

*Logged in as Member*
image

### Notes and questions
I also experimented with adding a link to the actual GitHub history for the document, which is possible to do if the `GITHUB_OWNER` and `GITHUB_REPO` environment variables are renamed to have the `TINA_PUBLIC` prefix so that the Tina client-side components can access them; obviously that introduces some amount of hassle in having to rename those variables across all of the existing sites, so I held off from including that in this PR pending confirmation of whether we really want/need that feature. Obviously it also opens up questions about whether Tina admins for a project will also have access to the GitHub repo, and how that will work logistically. (For the moment all of the content repos are public so it's not an issue, but that's something we will likely want to change going forward, since some clients might not want unpublished content sitting in a public repo...) Anyway, I can add those changes to this PR if desired.

## UPDATE 7/30/26
I've now also implemented an `editHistory` collection, which stores a record of every change made in any other collection, including renames and deletions. (Ignore the fact that the Name field is blank here; this data is from before I added that field, but it should correctly record the user's name now.)
image

There are definitely still improvements that could be made (table sorting/filtering is probably the lowest hanging fruit), but this should at least serve as an easily accessible log for admins to track.


tina/content/branding.ts
```diff
@@ -498,7 +516,7 @@ const Branding: Collection = {
         }]
       }]
     }]
-  }],
+  }, ...commonCollectionFields],
   ui: {
```
blms: Looks like `ui` keys duplicated in the same object, will that cause the audit log one to get overwritten? Any way to merge?

blms (commented):
Code looks good, just saw one place with a duplicated object key.

Otherwise, I'm wondering if we could include slightly more info in the log? Such as first name and last name from Clerk if it's available, and the individual fields that are edited? (The latter may not be as important here as in Fairdata since it's generally just pages with a few fields, but it may come in handy.)

Also, how big of a lift would it be to have a centralized audit log page like Camden has just added to FairData, so that we can also track page/post additions and deletions?

I guess the upside to appending user info to git commits is that they are already versioned diffs of the entire site with all change information. Otherwise I don't see much use in going the git route—especially as it adds a bit of a technical knowledge barrier to the end user for understanding git/github. So I think we probably don't need it if we can improve the in-Tina audit logging a bit.

tina/content/branding.ts
ajolipa: Oh good call, I just missed that it was already there; will fix!

netlify/functions/tina.ts
```diff
@@ -108,6 +108,10 @@ const ClerkBackendAuthentication = ({
           if (user.role === 'org:admin') {
             return { isAuthorized: true as const };
           }
+          // if this is an edit to the History collection, pass it through
+          if (req.body?.variables?.collection === 'editHistory') {
```
blms: Looking at the bodies passed below, is `collection` here ever going to be `editHistory` or is it going to be the collection name of the original record? Maybe I'm misreading. If I'm right maybe this could look for the mutation name instead of the collection name (like `DeleteDocument` below).

blms (approved):
Looks great! Just one question about this one auth check for non-admins.

blms (approved):
Looks good!

Issue #711, opened 2026-06-24
Keep the map working with no server


jamiefolsom:
**As a** project owner, **I want** the interactive map to keep working after the site is published as static files, **so that** visitors can explore places without me running a map server.

Part of #710, which covers static publishing.

PR #774, opened 2026-08-03
Fixing zoom behavior when `PRELOAD_MAP` is `true`

### LINKED ISSUE #736: Map doesn't zoom to search results when PRELOAD_MAP is on
With `PRELOAD_MAP=true`, the explore-page map stays at the world view instead of zooming to the results. The console shows `TypeError: Cannot read properties of undefined (reading 'getData')` from the map search code.

**Cause.** `getBoundingBox` in `src/apps/search/map/MapSearchContext.tsx` has two paths. Without preload it computes the box from the features it already has. With preload it calls `map.getSource('source-' + uuid).getData()` for every visible feature — assuming each feature's source is already on the map. The zoom effect can run before `MultiLayer` has added them all, so `getSource` returns undefined and the call throws an error. The error happens inside the Promise executor, so the promise never settles and nothing retries: the map sits at `0/0/0`.

Sites without `PRELOAD_MAP` never run this path, which is why most sites zoom fine.

**Suggested fix.** In the preload path, when `getSource` returns nothing, fall back to the feature itself (what the non-preload path does) or wait for the source to register — and make the promise settle on failure so the zoom can retry.

Seen on a static build (which requires `PRELOAD_MAP=true` for the geometry snapshot), but reproducible on any deployment with the flag on.

Part of #711. Every static build requires this setting, so every static site hits this bug.

ajolipa:
### In this PR
An attempt to address Issue #736  by a) preventing the `getData()` call on a source that's undefined; and b) if the first call to `getBoundingBox` fails to return any data because the map sources haven't loaded in yet, attempt it twice again at half second intervals and hope that's good enough.

### Notes and questions
Obviously this is not a very satisfying fix, and I have no idea what the right number of retries to go through is before declaring the source data a lost cause, but I don't want to just leave it in a loop of trying to set the bounding box if there's some issue fetching the data. Very much open to other approaches to this. It feels like there ought to be a way to get the `useEffect` hook to do the retrying for us as the map data fills in, but for reasons I don't entirely understand it only seems to trigger once before giving up. The console is littered with this particular error after the map data first loads, and I spent what started to feel like too much time trying to figure out exactly what's causing this: 
image
...but it seems to be something deep in the Peripleo package and have something to do with attaching mouseover event listeners? I'm not sure exactly how/whether this error relates to `getBoundingBox` failing to ever re-trigger, but the `setTimeout` thing is basically a way to circumvent the fact that it never naturally manages to try again if the first attempt at getting a bounding box returned no data.

### DIFF
--- a/src/apps/search/map/MapSearchContext.tsx
+++ b/src/apps/search/map/MapSearchContext.tsx
@@ -75,9 +75,10 @@ export const MapSearchContextProvider = ({ allowSave, children, preload }: Props
   /**
    * Returns a promise that resolves the bounding box for all visible features.
    */
-  const getBoundingBox = useCallback(() => new Promise<LngLatBoundsLike>((resolve) => {
+  const getBoundingBox = useCallback(() => new Promise<LngLatBoundsLike | null>((resolve) => {
     if (!map) {
-      return;
+      resolve(null);
+      return ;
     }
 
     const promises = [];
@@ -101,6 +102,8 @@ export const MapSearchContextProvider = ({ allowSave, children, preload }: Props
           promises.push(Promise.resolve(feature));
         } else if (source) {
           promises.push(source.getData());
+        } else {
+          promises.push(fetch(feature.properties.url).then((res) => (res.json())));
         }
       }
     });
@@ -110,7 +113,7 @@ export const MapSearchContextProvider = ({ allowSave, children, preload }: Props
         // Set the fetched data in the geometry cache
         setGeometryCache((prevCache) => ({
           ...prevCache,
-          ..._.indexBy(data, (d) => d.properties.uuid)
+          ..._.indexBy(_.compact(data), (d) => d.properties.uuid)
         }));
 
         // Calculate the bounding box
@@ -118,6 +121,9 @@ export const MapSearchContextProvider = ({ allowSave, children, preload }: Props
         const bbox = MapUtils.getBoundingBox(featureCollection);
 
         resolve(bbox);
+      })
+      .catch(() => {
+        resolve(null);
       });
   }), [features, geometryCache, map, preload]);
 
--- a/src/apps/search/map/MapSearchContext.tsx
+++ b/src/apps/search/map/MapSearchContext.tsx
@@ -86,7 +86,6 @@ export const MapSearchContextProvider = ({ allowSave, children, preload }: Props
     _.each(features, (feature) => {
       if (feature.properties.visible) {
         const id = feature.properties.uuid;
-        const source = map.getSource(`source-${id}`) as unknown as GeoJSONSource;
 
         const cached = geometryCache[id];
 
@@ -100,8 +99,6 @@ export const MapSearchContextProvider = ({ allowSave, children, preload }: Props
           promises.push(Promise.resolve(cached));
         } else if (!preload) {
           promises.push(Promise.resolve(feature));
-        } else if (source) {
-          promises.push(source.getData());
         } else {
           promises.push(fetch(feature.properties.url).then((res) => (res.json())));
         }
--- a/src/apps/search/map/MapSearchContext.tsx
+++ b/src/apps/search/map/MapSearchContext.tsx
@@ -99,7 +99,7 @@ export const MapSearchContextProvider = ({ allowSave, children, preload }: Props
           promises.push(Promise.resolve(cached));
         } else if (!preload) {
           promises.push(Promise.resolve(feature));
-        } else {
+        } else if (source) {
           promises.push(source.getData());
         }
       }
--- a/src/apps/search/map/MapView.tsx
+++ b/src/apps/search/map/MapView.tsx
@@ -36,27 +36,18 @@ const MapView = () => {
     !isRefinedWithMap() && route === '/' && config.map.zoom_to_place
   ), [route, isRefinedWithMap()]);
 
-  const fitMaptoBounds = useCallback((retries: number = DEFAULT_BOUND_RETRIES) => {
-    getBoundingBox().then((bbox) => {
-      if (bbox) {
-        map.fitBounds(bbox, boundingBoxOptions);
-      } else if (retries > 0) {
-        setTimeout(() => {
-          fitMaptoBounds(retries - 1);
-        }, 500);
-      }
-    })
-  }, [getBoundingBox, map, boundingBoxOptions]);
-
   /**
    * Sets the bounding box on the data set.
    */
   useEffect(() => {
     if (fitBoundingBox && !_.isEmpty(features) && map && !searching) {
-      fitMaptoBounds();
+      getBoundingBox().then((bbox) => {
+        if (bbox) {
+          map.fitBounds(bbox, boundingBoxOptions);
+        }
+      })
     }
   }, [boundingBoxOptions, fitBoundingBox, features, map, searching]);
-
 
   /**
    * Navigate to the `/select` route when feature is selected.
--- a/src/apps/search/map/MapView.tsx
+++ b/src/apps/search/map/MapView.tsx
@@ -6,8 +6,10 @@ import Map from '@components/Map';
 import { useGeoSearch, useSearching } from '@performant-software/core-data';
 import { useLoadedMap, useSelectionValue } from '@peripleo/maplibre';
 import { useCurrentRoute, useNavigate } from '@peripleo/peripleo';
-import { useContext, useEffect, useMemo } from 'react';
+import { useCallback, useContext, useEffect, useMemo } from 'react';
 import _ from 'underscore';
+
+const DEFAULT_BOUND_RETRIES = 3;
 
 const MapView = () => {
   const config = useSearchConfig();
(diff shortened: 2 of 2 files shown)


blms:
I agree that using retries sounds either brittle (could fail to fetch even when it should) or like it could potentially block rendering when there is genuinely nothing to fetch. Since this sounded like either a complex race condition or something deep in Peripleo (or both), I handed it off to Claude in the interest of saving time digging. Not sure if it's 100% accurate, like I don't know if the "partial sources" thing is a real state that could occur, but otherwise it seems pretty sound.

Apologies to just paste in Claude output as a PR review since that's not always terribly helpful, but here I think it genuinely may have surfaced the crux of the issue.

> The two changes are directionally right — the guard stops the crash and the retry usually lands — but it's a timeout papering over a race whose actual cause is identifiable and fixable. There's also one bug in the patch as written that will produce a *wrong* zoom rather than a missing one.
> 
> ## Why the source is missing (the part the issue doesn't name)
> 
> It's not that `MultiLayer` "hasn't gotten around to it yet" in a vague sense. `src/components/Map.tsx:126` gates all map children behind `WhenStyleLoaded`, which renders `null` until an effect + a `styledata`/`load` event flips `ready` (`src/components/Map.tsx:23`). So `MultiLayer` — and every `addSource` call — is deferred by at least one commit, often many frames.
> 
> `MapView`'s bbox effect is **not** behind that gate. Its guard is `map && features.length && !searching` (`src/apps/search/map/MapView.tsx:54`), and per the comment in `WhenStyleLoaded`, `useLoadedMap()` hands back the map as soon as a style prop exists, before the style parses. So the effect can fire in a commit where zero sources have been registered. And because nothing in `[boundingBoxOptions, fitBoundingBox, features, map, searching]` changes when the style finally loads and the layers mount, it never runs again — that's the "only triggers once" you couldn't explain. That gate was added in d37ff2d, which is itself a fix for map-search bounds.
> 
> Note this is a **gate**, not a gradual fill-in. Once `ready` flips, `addSource` is synchronous, so all sources appear in one commit. That matters: waiting on a real signal is reliable, and 500ms sampling is not the only option.
> 
> ## The console spam is unrelated — and it's a peripleo bug
> 
> `peripleo-maplibre.es2.js` is `GeoJSONLayer`. Decompiled, it does:
> 
> ```js
> useEffect(() => { map.addSource(`source-${id}`, { data }); ...addLayer x3; setSourceId(...) }, [map]);
> useEffect(() => { sourceId && map.getSource(sourceId).setData(data) }, [sourceId, data]);
> ```
> 
> `setSourceId` is state set *inside* the first effect, so the second effect always fires one render later and calls `setData` with data identical to what `addSource` just got. In maplibre 5.24, `loadData` opens with `this._pendingRequest?.abort()` — so that redundant `setData` cancels the in-flight fetch of the source URL, and the AbortError surfaces through `_dispatchWorkerUpdate`'s catch as an un-listened `error` event. One DOMException per layer, right after data first loads. Exactly what you're seeing.
> 
> This is why only preload sites see it: `SingleLayer` passes an inline object, so there's no network request to abort. `MultiLayer` passes `feature.properties.url` (`src/apps/search/map/MultiLayer.tsx:12`) — N layers, N aborted fetches. It's noisy and it restarts each load once, but it does not cause the bbox failure. Worth an upstream issue against `@peripleo/maplibre`, not something to fix here.
> 
> ## Bug in the patch: partial sources give a confidently wrong bbox
> 
> `else if (source)` silently drops features whose source is missing (`src/apps/search/map/MapSearchContext.tsx:102`). If even one source is registered, `Promise.all` resolves non-empty, `MapUtils.getBoundingBox` returns a valid box, `bbox` is truthy — so you `fitBounds` to a subset and never retry. The retry only saves you in the all-or-nothing case. Given the `WhenStyleLoaded` gate that's the common case today, but it's luck, and it fails silently (wrong zoom, clean console) rather than loudly.
> 
> The skip also poisons `geometryCache` less than you'd fear (it's keyed per uuid, so missing ones just aren't cached) — but the effect won't re-run to collect them.
> 
> ## The fix I'd actually make
> 
> **Stop using the map as the data source for the bounding-box calculation.** The map source and the bbox calculation both want the same bytes, and one of them is coupled to React/maplibre lifecycle for no reason. Each source URL is a prerendered single Feature with `properties.uuid` (`src/pages/api/geometry/[[uuid].json.ts`), and `feature.properties.url` is already on the feature. So in the preload branch:
> 
> ```diff
>    if (cached) {
>      promises.push(Promise.resolve(cached));
>    } else if (!preload) {
>      promises.push(Promise.resolve(feature));
> -  } else if (source) {
> -    promises.push(source.getData());
> +  } else {
> +    promises.push(fetch(feature.properties.url).then((r) => r.json()));
>    }
> ```
> 
> 
> Same URL the layer fetches, so it's an HTTP cache hit; `fetchGeometry` in `src/backend/api/geometry/index.ts` already does this and is used by `src/visualizations/Map.tsx:47`. No race, no retries, no dependency on layer mount order, and the existing per-uuid `geometryCache` keeps it to one fetch per layer — which was the whole reason `getData()` was chosen. Delete the retry ladder entirely.
> 
> **Better still, if you want to close this for #711 properly:** precompute the bbox at build time. `getAll()` in `src/services/geometry.ts` already has `data.geometry` in hand, so `src/pages/api/geometry/index.json.ts` can emit `{ id, url, bbox }` per record, and the overall bbox becomes a pure reduce over `geometries` — zero network I/O, zero map coupling, instant zoom on static sites. That's a slightly larger change but it makes the preload path *faster* than the non-preload path instead of more fragile.
> 
> ## Fixes worth making regardless of which approach you pick
> 
> - `src/apps/search/map/MapSearchContext.tsx:79` — `if (!map) return;` inside the executor leaves the promise permanently pending. Return `Promise.resolve(null)` before constructing it.
> - `src/apps/search/map/MapSearchContext.tsx:108` — `Promise.all` has no `.catch`. Any rejection strands the promise. `.catch(() => resolve(null))`.
> - `src/apps/search/map/MapSearchContext.tsx:113` — `_.indexBy(data, (d) => d.properties.uuid)` throws if anything resolves without `.properties` (a null, or a FeatureCollection once `cluster_radius` is set). Compact first.
> - If you do keep a retry: make it all-or-nothing. Count expected-vs-resolved and `resolve(null)` on a shortfall, so a partial result retries instead of zooming wrong.
> - `src/apps/search/map/MapView.tsx:39` — `fitMaptoBounds` is called from the effect but omitted from its deps; adding it would loop, because `getBoundingBox` depends on `geometryCache`, which `getBoundingBox` itself sets. That circular dep is the structural reason this hook is hard to retry cleanly, and it disappears if geometry fetching moves out of the bbox function.
> - The `setTimeout` isn't cleared on unmount — it can `fitBounds` on a torn-down map during route changes.
> - Typo: `fitMaptoBounds` → `fitMapToBounds`; stray double blank line at `src/apps/search/map/MapView.tsx:58`.
> 
> ## Two things to check that I couldn't confirm statically
> 
> `MultiLayer` returns an unkeyed `<>` per item from `_.map` — the keys are on the children *inside* the fragment, so React reconciles those layers by index. Today `getFeatures` only ever appends (it marks removed items `visible: false` rather than splicing), so indices stay stable and it works by accident. But `GeoJSONLayer`'s add-source effect has deps `[map]` and captures the source id at mount, so if the array order ever changes, a layer keeps its old source id while receiving a new `layerId` — `getSource('source-' + newUuid)` would then return undefined *permanently*, not transiently. Worth switching to `<Fragment key={uuid}>` prophylactically.
> 
> Second: I traced this from source rather than reproducing it. Confirming the ordering is cheap — log in both `WhenStyleLoaded`'s `setReady` and the `MapView` effect on a `PRELOAD_MAP=true` build and check that the effect fires first. If it does, the `WhenStyleLoaded` diagnosis is confirmed and the direct-fetch fix is unambiguously correct.

ajolipa:
Implemented the suggestion to fetch the geometry directly from the API in the case that the source is not found, and it seems to work in testing. Yay! I was hesitant to add a duplicate API call since sites with the geometry preloaded will be ones with more volume of map data, but I tried it locally with PADP and the performance seemed okay. (I also don't know exactly how `getData` was working under the hood anyway; this might not be any worse in terms of how the data is moved around.)

Also did some of the smaller fixes as suggested, allowing the promise to resolve to `null` in the case of error or a missing map.

src/apps/search/map/MapView.tsx
```diff
@@ -6,9 +6,11 @@ import Map from '@components/Map';
 import { useGeoSearch, useSearching } from '@performant-software/core-data';
 import { useLoadedMap, useSelectionValue } from '@peripleo/maplibre';
 import { useCurrentRoute, useNavigate } from '@peripleo/peripleo';
-import { useContext, useEffect, useMemo } from 'react';
+import { useCallback, useContext, useEffect, useMemo } from 'react';
 import _ from 'underscore';
 
+const DEFAULT_BOUND_RETRIES = 3;
```
blms: Should we remove `useCallback` and `DEFAULT_BOUND_RETRIES` now?

src/apps/search/map/MapSearchContext.tsx
```diff
@@ -99,8 +100,10 @@ export const MapSearchContextProvider = ({ allowSave, children, preload }: Props
           promises.push(Promise.resolve(cached));
         } else if (!preload) {
           promises.push(Promise.resolve(feature));
-        } else {
+        } else if (source) {
           promises.push(source.getData());
```
blms: Should we remove this `else if (source)` path entirely now? Based on Claude's analysis of it, it sounds like this would still introduce the unbounded promise when a source exists but its URL fails due to a network or server error. Here's [getData](https://github.com/maplibre/maplibre-gl-js/blob/3d50bcfd22859f81df16b1859e46c8445495e6e9/src/source/geojson_source.ts#L303) fwiw.

blms (approved):
Looks good!

Issue #589, opened 2026-03-24
Define RBAC test checklist for TinaCMS and Astro upgrades


jamiefolsom:
The role-based UI restrictions in PR #586 depend on TinaCMS's internal DOM structure via CSS selectors and MutationObserver. These could break silently when upgrading TinaCMS or Astro.

Define a manual or automated test checklist to run after upgrades:

**Sidebar locking (editor role):**
- [ ] Settings, Branding, Internationalization, Navbar, Pages show as locked with lock icon
- [ ] Locked links are not clickable
- [ ] Posts, Paths remain accessible

**Form read-only (editor viewing non-owned content):**
- [ ] Ownership banner appears
- [ ] All text fields disabled (pointer-events: none)
- [ ] Date picker disabled
- [ ] Image dropzone disabled
- [ ] Rich text editor disabled
- [ ] Save button disabled
- [ ] Status dot replaced with lock icon

**Form editable (editor viewing own content):**
- [ ] No banner
- [ ] All fields interactive
- [ ] Save works

**Admin override:**
- [ ] No sidebar restrictions
- [ ] All content editable regardless of owner

**Non-Clerk sites:**
- [ ] No restrictions visible
- [ ] All collections and content accessible

CSS selectors to watch: `.relative.w-full.flex-1.overflow-hidden` (form scroll area), `.border-b.border-gray-100` (form header bar with status dot).

Consider automating with Playwright using the `TINA_PUBLIC_DEV_ROLE` override for role switching.

jamiefolsom:
Pull request #777 added the tests this checklist asked for and was merged on August 11. Closing.

Issue #587, opened 2026-03-24
Pre-existing content without creator field is editable by all editors


jamiefolsom:
Posts and paths created before the Clerk RBAC feature (PR #577) have no `creator.id` field. The `OwnershipNotice` component in PR #586 treats content without a creator as owned by the current user, making it editable by any editor.

**Options:**
1. Treat unclaimed content as admin-only (editors can view but not edit)
2. Run a migration to assign `creator` fields to existing content based on git history or a manual mapping
3. Accept the current behavior (first editor to save claims ownership via `beforeSubmit` auto-population)

Needs a policy decision. Related to the broader Clerk migration effort.

PR #777, opened 2026-08-05
Adding basic tests for non-admin Tina UI changes

ajolipa:
### In this PR
As discussed in Issue #589 , the current way that we're applying UI changes to the Tina admin interface based on the permissions of the authenticated user is potentially brittle in the face of potential updates to the exact DOM/CSS structure of the default Tina UI. After any version updates or other changes to how the Tina UI functions, we want the ability to make sure that this functionality hasn't broken. This PR introduces some first-pass checks to make sure that sidebar links and editor fields are being correctly modified according to the styles in the `/tina/role-ui.ts` file.

### Notes
- This set of tests does not yet include explicitly checking that admin users can access everything. To me that part of the functionality seems more robust and unlikely to break in subtle ways after an update so it didn't really feel necessary, and didn't feel worth the time to try to figure out the best way to change the role on the fly. 
- To run the tests, you must have a testing environment set up with `TINA_PUBLIC_IS_LOCAL=true` and `TINA_PUBLIC_DEV_ROLE=org:member`. Additionally the testing site must have at least one post and one path, and it must be the case that the first post and first path on the lists have a `creator` listed. (Otherwise, as noted in Issue #587 , that first entry will in fact be editable by any authenticated user.)
- It may be overkill to even have a Playwright test for this, since a human can just open the testing environment and inspect the relevant elements at least as quickly as the automated test can, and it's not something that needs to be tested separately for every deployed site. But it also can't hurt to have the automated test available, I suppose.


README.md
```diff
@@ -36,11 +36,20 @@ npm run vitest
...
 ```
-npm run playwright
+npm run test-a11y
 ```
 
 The results will be output to `playwright-report/index.html`.
 
+#### RBAC tests
+
+To test that RBAC is being correctly enforced in the TinaCMS interface, set the `TINA_PUBLIC_TINA_BASE_URL` to the admin URL you want to test (or leave blank to default to `http://localhost:8888/admin`), and set `TINA_PUBLIC_IS_LOCAL=true` and `TINA_PUBLIC_DEV_ROLE=org:member` in the environment you wish to test. Then run:
```
blms: Should this `TINA_PUBLIC_TINA_BASE_URL` be called something more like `RBAC_BASE_URL` since it's not something exposed to the client and is specific to RBAC testing? 

Also, can the new env vars be added to `.env.example`? I don't see the new one from this PR, or `TINA_PUBLIC_DEV_ROLE` or `TINA_PUBLIC_DEV_USER_ID` in there.

Finally, going by the tests it looks like the `/admin` might need a trailing slash like `/admin/`.

test/browser/rbac.test.ts
```diff
@@ -0,0 +1,66 @@
...
+  });
+
+  test('Read-only Posts editor for non-admin user', async ({ page }) => {
+    test.setTimeout(60_000)
+    await page.goto('index.html#/collections/post');
+    await page.locator('td a').first().click();
+
+    await expect(
+      page.locator(
+        `text=You're not the author of this content. You can view but not edit.`
```
blms: This seems brittle in that it expects the first post to be one where this is the case, and it seems to require a bit of an onerous setup on the part of the developer.

Could we include a minimal fixture set—where we can guarantee the necessary preconditions—and test against that instead of live content produced by the dev? Like a pre-populated `content` directory, something like `test/fixtures/content` that has the exact expected posts and paths etc, and then point the tests at it via `TINA_LOCAL_CONTENT_PATH` only when running tests.

e.g., two terminals running
```sh
TINA_PUBLIC_IS_LOCAL=true TINA_PUBLIC_DEV_ROLE=org:member TINA_PUBLIC_DEV_USER_ID=user_test_fixture TINA_LOCAL_CONTENT_PATH=test/fixtures/content npm start
```
and
```sh
TINA_PUBLIC_TINA_BASE_URL=http://localhost:4321/admin/ npm run test-rbac
```

blms (commented):
I'm wondering if we can run this in github actions CI on every pull request as opposed to having the dev manually run test suites? At least I'm probably not going to remember to run them before opening a PR every time…

I had claude spit out a quick workflow, probably not usable exactly as written but could provide the basic scaffolding:

<details><summary>github workflow</summary>

```yml
name: Test

on:
  pull_request:
  push:
    branches: [develop]

concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true

permissions:
  contents: read

env:
  TINA_LOCAL_CONTENT_PATH: test/fixtures/content

jobs:
  unit:
    name: Unit tests
    runs-on: ubuntu-latest
    timeout-minutes: 10
    steps:
      - uses: actions/checkout@v4

      - uses: actions/setup-node@v4
        with:
          node-version-file: .node-version
          cache: npm

      - run: npm ci

      # Generates public/config.json, public/branding.json, src/i18n/*.json
      # and src/components/custom/project — all gitignored, all required by
      # the vitest path aliases (@config, @branding).
      - name: Build generated assets
        run: node scripts/build.mjs

      - run: npm run vitest

  rbac:
    name: TinaCMS RBAC
    runs-on: ubuntu-latest
    timeout-minutes: 20
    env:
      TINA_PUBLIC_IS_LOCAL: 'true'
      TINA_PUBLIC_DEV_ROLE: 'org:member'
      TINA_PUBLIC_DEV_USER_ID: 'user_test_fixture'
      TINA_PUBLIC_TINA_BASE_URL: 'http://localhost:4321/admin/'
    steps:
      - uses: actions/checkout@v4

      - uses: actions/setup-node@v4
        with:
          node-version-file: .node-version
          cache: npm

      - run: npm ci

      - run: npx playwright install --with-deps chromium

      # npm start === node scripts/build.mjs && tinacms dev -c "astro dev".
      # The TINA_PUBLIC_* vars above are inlined into the admin bundle here,
      # which is why they must be set on this step and not on the test step.
      - name: Start TinaCMS admin
        run: nohup npm start > /tmp/dev-server.log 2>&1 &

      - name: Wait for admin
        run: |
          for i in $(seq 1 90); do
            curl -sfo /dev/null http://localhost:4321/admin/index.html && exit 0
            sleep 2
          done
          echo "::error::admin did not come up within 180s"
          cat /tmp/dev-server.log
          exit 1

      - run: npm run test-rbac

      - name: Upload report
        if: failure()
        uses: actions/upload-artifact@v4
        with:
          name: playwright-report
          path: playwright-report/
          retention-days: 7

      - name: Dump server log
        if: failure()
        run: cat /tmp/dev-server.log
```
</details>

blms (approved):

PR #779, opened 2026-08-10
Adding some documentation for creating new Tina collections

### LINKED ISSUE #590: Document RBAC requirements for adding new TinaCMS content types
When new TinaCMS collections are added to CDP that editors should be able to access (like Posts and Paths), the RBAC system requires specific setup. This should be documented as part of the "add a new collection" checklist.

**For editor-accessible collections, add:**
1. \`_ownershipNotice\` pseudo-field (using \`OwnershipNotice\` component) as the first field
2. \`creator\` object field with \`id\` and \`email\` sub-fields (using \`Creator\` component)
3. \`beforeSubmit\` hook with ownership guard (\`getUserRole\` check) and creator auto-population
4. Ensure the collection name is NOT in \`ADMIN_ONLY_COLLECTIONS\` in \`tina/role-ui.ts\`

**For admin-only collections:**
1. Add the collection label to \`ADMIN_ONLY_COLLECTIONS\` in \`tina/role-ui.ts\`
2. No ownership fields needed

See \`tina/content/posts.ts\` as the reference implementation for editor-accessible collections.

ajolipa:
### In this PR
Adds some basic documentation outlining the steps for adding a new collection to Tina, both in the case of an admin-only collection and a collection (like Paths and Posts) for member-generated content. This should hopefully help us make sure that we dot all the i's etc when setting up new content so that the RBAC works as expected.

Resolves #590 .

### DIFF
--- a/docs/adding-content-collections.md
+++ b/docs/adding-content-collections.md
@@ -1,0 +1,144 @@
+# Adding a New Tina Collection
+
+Developers should use the following steps to add a new type of content to the TinaCMS schema. The steps differ based on whether the new collection should be accessible to editors or only to site admins.
+
+## Common Steps
+1. Create a file `newCollectionName.ts` in `tina/content` (e.g. `stories.ts`).
+2. In your new file, define your `Collection` following this example (obviously replacing `stories` with your new collection name):
+```
+import { Collection } from '@tinacms/schema-tools';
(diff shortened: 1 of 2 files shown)


blms (approved):
Looks great and easy to read!

PR #780, opened 2026-08-10
RBAC tests revisions

blms:
### In this PR

Here's a PR against your branch—I had to get Claude and Gemini to fight each other and do a bit of research myself to get this working 😅 but it seems to be working now, at least locally.

- The main issue: the test runner nested `astro dev` inside `netlify dev`, which would fail silently with exit code 126 ("command found but cannot execute")
   - It's resolved by separating them such that Astro runs directly, and Netlify runs as a proxy-only server in front (for its edge functions runtime)
   - This is done in a new config file `playwright.rbac.config.ts`, which starts the two servers, injects the hard coded test env vars, and sets up the environment for serial tests via chromium
   - Semi-related: test suite now waits for warmup in `beforeAll` so that the server is fully ready when tests run; and `netlify-cli` is added as a dev dependency to ensure everything is running under the same Node version
- A couple of test code changes: ensure the "Enter edit mode" button is pressed correctly; ensure "Edit History" does not match on "Edit History Dashboard" with `exact`
- Derive test fixture path from `TINA_LOCAL_CONTENT_PATH` rather than having a second variable
   - New `scripts/build.paths.mjs` — a small helper that derives the content directory from `TINA_LOCAL_CONTENT_PATH` as a function, which then gets reused in the two other build scripts to build the content path
   - **Note:** Now, when `TINA_LOCAL_CONTENT_PATH` is set, content is read locally and not fetched from GitHub, even if `GITHUB_REPO` and `GITHUB_OWNER` are set. LMK if this should not be the case.

Now let's see if it works under CI… success!

#### Note: This also broke local development using `netlify dev` under node 24

The new script for local development is:
```
npm install && npm run dev
```

which uses `concurrently` to run both servers and `wait-on` to wait for the astro server to start before starting netlify.


blms:
Converting this back to a draft because it seems to break running `netlify dev` locally, as you mentioned in standup under node 24. It seems to be an issue with netlify trying to auto-integrate with astro and that particular path breaking under node 24. Working on a fix now

EDIT: Fixed, using a new script called `npm run dev` and a couple of dev dependencies to get the waiting and concurrent processes working

playwright.rbac.config.ts
```diff
@@ -0,0 +1,96 @@
...
+  projects: [{
+    name: 'rbac',
+    use: {
+      ...devices['Desktop Chrome']
+    }
+  }],
+
+  // Playwright starts both and waits for both; netlify's --target-port wait
+  // bridges the gap until astro binds 4321.
+  webServer: isLocalAdmin ? [
```
ajolipa, line 55: Just making sure I understand the flow here -- suppose I want to run the test on my own machine against and already-running `localhost`? Does this config allow for that or will it automatically attempt to spin up the local server from scratch? Maybe not an important enough use case to worry about but I can imagine wanting to run tests as part of doing other dev work and it would be nice if this setup played well with that.

playwright.rbac.config.ts
```diff
@@ -0,0 +1,96 @@
...
+    ['list'],
+    ['html', { open: 'never' }]
+  ],
+
+  use: {
+    baseURL,
+    trace: 'on-first-retry',
+  },
+
+  // The admin UI is Chromium-only in practice.
```
ajolipa: Is there a specific reason not to test other browsers here? Why is the admin UI Chromium-only in practice?

netlify.toml
```diff
@@ -14,8 +14,10 @@
   force = true
 
 [dev]
-  command = "npm start"
+  # A dummy command to keep the Netlify process alive without spawning Astro
+  command = "tail -f /dev/null"
```
ajolipa: This seems to interfere with the normal functionality of running `netlify dev` with a linked project -- injecting environment variables, etc. I just tried to spin up a local instance of the PSS Scavenger Hunt site and it failed to actually grab any content. Could we use a `context` here to allow flags that control whether `netlify dev` runs this version or just `npm start`? E.g. add a context block:
```
[context.testing]
command = "tail -f /dev/null"
```
And then add the `--context=testing` flag to the `proxy` command in `package.json`. I think something like that would allow us to have different setups for local dev and for testing.

playwright.rbac.config.ts
ajolipa: Would changing `reuseExistingServer` in this config to `!!process.env.CI` possibly fix this? Or does the `--kill-others` flag in the `dev` command just make the `reuseExistingServer` flag moot?

ajolipa, line 55: From what I can tell it's supposed to be the case that Playwright will clean up any servers that were started for its tests, but when I just ran these tests locally via VSCode I notice that the site is still being served on port 8888 after the test has finished. With all of the proxying etc do we have to do something more explicit to make sure that everything is shut down properly when the test is done?

ajolipa (commented):
This is amazing and you're a superhero! I had a couple of questions, mostly to do with how this interacts with existing local dev setup, but it seems that the CI tests were able to complete successfully, so this definitely seems like a successful approach.

playwright.rbac.config.ts
blms: Good point, I was thinking that we don't really need that setup because tests run against only the fixture content, not really the user's live content. But I guess we would want that scenario when developing new tests and fixtures?

Apparently playwright has a "watch" mode that will allow you to do that, editing fixtures/tests with a live server:
```
npx playwright test --config=playwright.rbac.config.ts --ui
```

If that works, it may be preferable to allowing the tests to run against arbitrary live content.

(fwiw `--kill-others` is unrelated, just ensures that astro and netlify servers both die together)

playwright.rbac.config.ts
blms: I think that's an LLM hallucination, not sure what it means. Removed the comment. We can test against other browsers, though it will increase the test suite's runtime, let me know if we should!

netlify.toml
blms: So this was actually an intentional change, for dev and not just for testing, because `netlify dev` was failing with the same timeout in my dev environment under node 24 (see https://github.com/performant-software/core-data-places/pull/780#issuecomment-5244346966 above).

I made the new script `npm run dev`, which should correctly run both servers independently (you'll need to `npm install` first). 

If we need to still be specifically using `netlify dev` for some reason (and/or it already works for you under node 24!) then let me know, and I can change the flow a bit such that the no-op thing happens only when running `npm run dev`, and `netlify dev` will still run the command as before.

playwright.rbac.config.ts
blms: Good point! I added a global teardown step to make sure all processes are ended.

netlify.toml
ajolipa: I agree it would be great to solve the timeout problem more robustly. Is there some way to use the new setup along with the Netlify CLI functionality of injecting environment variables? It's a bit of a hassle to switch back and forth between projects if I have to manually change my own `.env` file, is the only issue.

playwright.rbac.config.ts
ajolipa: I think testing on any one browser should be sufficient in this case; since we're really just examining things in the DOM and making sure elements have been modified as expected, I would be surprised if we ran into any browser-specific failures. So I'm happy leaving it to just Chromium; I was just thrown off by the comment haha.

netlify.toml
blms: Ah, ok I didn't know the netlify CLI did that! Let me look into this a bit more 👍

blms: Ok, I've updated the `npm run dev` script to use `netlify dev:exec` ([docs](https://cli.netlify.com/commands/dev/#devexec)) which supposedly will pull in env vars from a linked site. Let me know if it doesn't work.

ajolipa: This does work! Good call.

playwright.rbac.config.ts
ajolipa: Hmm, when I try this with an already-running dev server it gives me the error http://localhost:4321 is already in use. Make sure that nothing is running on the port/url or set `reuseExistingServer: true in config.webServer.` But I agree that at least with these particular tests it probably isn't necessary to run them against arbitrary content. I could imagine other test suites that we'd want the ability to test against a variety of different user-generated content, but we can cross that bridge when we come to it.

scripts/build.content.mjs
```diff
@@ -1,14 +1,23 @@
...
 import fs from 'node:fs';
+import { contentPath } from './build.paths.mjs';
 
 const TEMP_DIR = './tmp';
-const LOCAL_CONTENT_ROOT = process.env.BASE_LOCAL_CONTENT_PATH || './';
 
 export const fetchContent = async () => {
-  if (!(process.env.GITHUB_OWNER && process.env.GITHUB_REPO)) {
+  // Content already exists locally; don't fetch (cloning would overwrite it).
+  const useLocalContent = !!process.env.TINA_LOCAL_CONTENT_PATH;
```
ajolipa: Not sure I like this; when I'm developing locally I often want to be running Tina in local mode (so that I don't have to worry about auth and changes are only saved locally) but I still want to fetch content initially from whatever site I'm currently using to test. I know it's annoying to just keep adding flags but could `useLocalContent` be controlled separately from just whether or not you have the Tina content path set?

ajolipa (commented):
Super close to covering every base I can think of; just one more question about how the modification to the build script interacts with local dev workflow.

playwright.rbac.config.ts
blms: Oops, I was unclear, that shouldn't be run at the same time as a separate live dev server! Instead it spins up its own live server with fixture data loaded that you can use to edit both tests and fixture data. Seemed to work for me by running `npm run test-rbac -- --ui`.

But yeah this is only for fixture data and not arbitrary content.

scripts/build.content.mjs
blms: I think I get what you mean—something like this? e49c696

ajolipa (approved):
Looks good!

PR #781, opened 2026-08-12
Allow port overrides in the dev scripts

jamiefolsom:
### In this PR

The dev scripts accept optional port overrides through environment variables: `FDS_ASTRO_PORT`, `FDS_NETLIFY_PORT`, `FDS_GRAPHQL_PORT`, and `FDS_DATALAYER_PORT`. The defaults (4321, 8888, 4001, 9000) are unchanged, so nothing changes unless you set them.

### Why

#777 and #780 set the right direction: the repo itself now knows how to start its own dev servers reliably, in CI and locally. The pstudio CLI is adopting that same workflow as the standard way to run a site. When the checked-out version has `npm run dev`, pstudio will run it instead of maintaining its own copy of the startup logic; for older versions that don't have it, pstudio keeps a fallback. pstudio runs each site in its own workspace on its own ports so several sites can run at once — that's what the overrides are for.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,8 @@ npm install && npm run dev
 
 `npm run dev` runs Astro and a Netlify proxy as separate processes (Netlify spawning Astro breaks under Node 24). Astro runs via `netlify dev:exec`, so a linked Netlify site's environment variables are injected.
 
+The ports default to 4321 (Astro), 8888 (Netlify), 4001 (TinaCMS GraphQL), and 9000 (TinaCMS datalayer). To run more than one site at a time, override them with the environment variables `FDS_ASTRO_PORT`, `FDS_NETLIFY_PORT`, `FDS_GRAPHQL_PORT`, and `FDS_DATALAYER_PORT`.
+
 ## Testing
 
 #### Unit tests
--- a/package.json
+++ b/package.json
@@ -3,8 +3,8 @@
   "version": "0.0.1",
   "private": true,
   "scripts": {
-    "start": "node scripts/build.mjs && tinacms dev -c \"astro dev\"",
-    "proxy": "wait-on tcp:4321 && netlify dev --offline -c \"tail -f /dev/null\"",
+    "start": "node scripts/build.mjs && tinacms dev -p ${FDS_GRAPHQL_PORT:-4001} --datalayer-port ${FDS_DATALAYER_PORT:-9000} -c \"astro dev --port ${FDS_ASTRO_PORT:-4321}\"",
+    "proxy": "wait-on tcp:${FDS_ASTRO_PORT:-4321} && netlify dev --offline -c \"tail -f /dev/null\" --port ${FDS_NETLIFY_PORT:-8888} --target-port ${FDS_ASTRO_PORT:-4321}",
     "dev": "concurrently --kill-others --names \"ASTRO,NETLIFY\" \"netlify dev:exec npm start\" \"npm run proxy\"",
     "build": "node scripts/build.mjs && npm run test-config && tinacms build && astro build",
     "preview": "astro preview",


blms:
In general, makes sense to me to make these ports configurable.

Used Claude to review for any potential downstream issues encountered by setting these, here's what it said:

> The change itself is small and correct: `package.json`'s `start`/`proxy` scripts now take port overrides via `${FDS_*:-default}` shell expansion (valid — npm runs these through `sh`), and the flags used (`astro --port`, `netlify --port/--target-port`, `tinacms -p/--datalayer-port`) are all real. Three of the four ports wire up cleanly. But there's one real bug and a couple of downstream landmines.
> 
> ### 🔴 The datalayer override is only half-wired (functional bug)
> 
> The script moves the datalayer **server** with `--datalayer-port ${FDS_DATALAYER_PORT:-9000}`, but the datalayer **client** in `tina/database.ts:21` reads a *different* variable, `TINA_DATALAYER_PORT`. Nothing bridges the two.
> 
> The code comment right there warns exactly about this:
> > Tina's `--datalayer-port` flag moves only the server; this client must follow, or local dev hangs at "Indexing local files".
> 
> So a developer who follows the new README and sets **only** `FDS_DATALAYER_PORT=9001` to run a second site moves the server to 9001 while the client still hits 9000 → dev hangs at "Indexing local files." The README (`README.md:28`) actively instructs this broken path.
> 
> Fix options (either works):
> - Have the client fall back: `process.env.TINA_DATALAYER_PORT ?? process.env.FDS_DATALAYER_PORT` in `tina/database.ts`, **or**
> - Export the mapping in the `start` script: prefix with `TINA_DATALAYER_PORT=${FDS_DATALAYER_PORT:-9000}` so the client follows automatically.
> 
> I'd lean toward the script-level mapping so `FDS_*` stays the single public knob.
> 
> ### 🟠 Downstream: RBAC Playwright config hardcodes 4321/8888
> 
> `playwright.rbac.config.ts` launches `npm start` as its web server — which now honors `FDS_ASTRO_PORT` — but hardcodes the port it waits on (`port: 4321`, line 60), the proxy target (`--target-port 4321`, line 16), and `baseURL` (`:8888`, line 6). `test/browser/rbac.teardown.ts:7` also frees `[8888, 4321]` literally.
> 
> Consequence: if a developer has `FDS_ASTRO_PORT`/`FDS_NETLIFY_PORT` set in their `.env`, `npm run test-rbac` will start Astro on the overridden port while Playwright waits on 4321 → timeout. Clean CI is unaffected (no overrides), so this is a local-dev footgun rather than a CI break — but it's the most likely place the new vars silently bite someone.
> 
> ### 🟡 Minor / stale, not broken
> - **netlify.toml [dev]** still hardcodes `targetPort = 4321` / `port = 8888` (`netlify.toml:18`). Functionally harmless — the `proxy` script passes explicit `--port`/`--target-port` flags that override the TOML — but it's now inconsistent with the overridable scripts.
> - **`PUBLIC_BASE_URL`** in `.env.example` is `http://localhost:8888/` and is consumed by `src/utils/galleries.ts:3`. Overriding `FDS_NETLIFY_PORT` without updating a developer's `.env` would leave this stale. It's `.env`-controlled, so acceptable, but worth a note.
> 
> ### ✅ Fine as-is
> `FDS_GRAPHQL_PORT` (the Tina GraphQL server) is safe to move: the app reaches Tina through the relative `contentApiUrlOverride: '/api/tina/gql'` (`tina/config.ts:47`) → Netlify proxy → the `tina` function, not via a hardcoded `4001`. Astro and Netlify port overrides work as intended.
> 
> ---
> 
> **Bottom line:** the Astro/Netlify/GraphQL overrides are good, but the datalayer override doesn't actually work end-to-end and the README promises it does — that's worth fixing before merge. The Playwright hardcodes are the main downstream surprise.

PR #783, opened 2026-08-13
Change `sanitize-html` back to v2.17.1

ajolipa:
### In this PR
Attempt to address issue where Netlify SSR is throwing errors when trying to load the `sanitize-html` package because of the presence of a commonJS import.

Tested in a branch deploy with the JUEL documents and the fix does seem to work: https://rb-update-sanitize-html-version--juel-staging.netlify.app/en/items/0aa3d451-f519-43cc-ac11-d5a4895ad603

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ npm install && npm run dev
 
 ## Testing
 
-#### Unit tests
+#### Unit tests 
 
 ```
 npm run vitest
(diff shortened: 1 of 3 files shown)


blms (approved):

jamiefolsom (approved):
Go for it!

camdendotlol (approved):

